PR #1, opened 2010-09-16
HTTP::Config handler priority

frett:
this is a patch that adds support for specifying a custom priority for items in a HTTP::Config object.

I need this support to be able to attach several response_done handlers that need to fire in a specific order.

here is the RT for this issue: https://rt.cpan.org/Public/Bug/Display.html?id=59497

### DIFF
--- a/lib/HTTP/Config.pm
+++ b/lib/HTTP/Config.pm
@@ -423,6 +423,12 @@ Matches if either the request or the response have a header $field with the give
 
 Matches if the response object has that key, or the entry has the given value.
 
+=item m_priority => $number
+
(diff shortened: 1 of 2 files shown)


genio:
@frett Can you update your PR with a pull and rebase so we can revisit this one, please?

genio:
Hi, @frett 

I'm closing this for now because ```HTTP::Config``` is now part of the [HTTP::Message](https://github.com/libwww-perl/HTTP-Message) distribution.  Please create the PR in that dist if you still want the feature and we'll take a look.

Sorry for the inconvenience!

Thanks,
Chase

PR #23, opened 2011-12-03
LWP::UserAgent: Chainable methods

dolmen:
I've modified the methods that do not return a meaningful value to instead return $self.
This allows to chain some methods calls: add_handler, set_my_handler, env_proxy, no_proxy.


dolmen:
@genio 5 years later I still thinks this feature is valuable. It allows to instantiate an LWP::UserAgent and call its `env_proxy` method in a single expression.

PR #28, opened 2012-04-23
Fix many tests using ok() instead of like()

schwern:
t/local/http.t has many tests of the form:

    ok $res->content, qr/foo/;

which does nothing but check `$res->content` is true.  They should be using `like`.


karenetheridge:
Thanks, merged!

PR #31, opened 2012-11-08
Use ->content as backup plan when ->decoded_content is undef.

gslin:
->decoded_content might be undef when charset unknown, we'll use ->content as backup plan.

### DIFF
--- a/lib/LWP/RobotUA.pm
+++ b/lib/LWP/RobotUA.pm
@@ -132,7 +132,7 @@ sub simple_request
 	my $fresh_until = $robot_res->fresh_until;
 	my $content = "";
 	if ($robot_res->is_success && $robot_res->content_is_text) {
-	    $content = $robot_res->decoded_content;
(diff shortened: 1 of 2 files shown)


simbabque:
I don't think we should do this change. In HTTP::Message, which implements `decoded_content`, there isn't a credible use case for returning `undef` from it [in the unit tests(https://github.com/libwww-perl/HTTP-Message/blob/master/t/message.t#L433)] apart from the Message not being processable. It's very clearly documented that it is supposed to fail if it cannot decode the content, and it will set `$@`.

This change would break the contract on that, and change behaviour that we've had for a long time. `decoded_content` is `undef` for a reason, and that reason should be checked and handled in application code.

genio:
Given the reasons @simbabque lined out, I don't think we can accept this PR. If you'd like to talk it out, please join us on IRC and we'll be happy to go over scenarios. For now, though, I think I'm going to close this one out.

Thank you

Issue #39, opened 2013-04-20
Request with digest authentication fails if nonce is expired


marmotil:
There is a problem with digest authentication together with any request, e. g. GET, when the server expires the nonce. In this case, the request finally fails with status 401 instead of retrying with the new nonce received from the server.

This always happens if the UserAgent object is used longer than the expiry of the nonce while querying the same server, e. g. when used in mod_perl to query a different URL. The server is free to expire the nonce at any time, the default setting on Apache is after 300 s.

As the authenticate() method is inherited in LWP::Authen::Digest from LWP::Authen::Basic, this case was not foreseen: With basic authentication, if the credentials were already sent to the server for the realm and were rejected, this is always a final failure. With digest authentication, this is also true, except when the nonce is expired, indicated by the server by the "stale=true" authentication parameter.

In this case, the authentication should be repeated, like when a 401 status is received and the authentication parameters were not yet sent to the server.

oalders:
Re-opening as I think it's helpful to have an open issue in addition to the open PR

PR #40, opened 2013-04-20
Fix digest authentication failure if nonce is expired

marmotil:
This change fixes the digest authentication failure described in issue #39.

An additional method reauth_requested() is introduced to the authentication mechanism. It shall indicate that the server suggests a reauthentication even if the authentication attempt fails.

It is always false for the basic authentication and all classes inheriting from that, unless they override it. For the digest authentication, it also returns false, unless the nonce is expired, indicated by the "stale=true" authentication parameter. In this case, the stale nonce from the request is deleted from the history, and the request is retried.

Tests were added to check this new behavior.

### DIFF
--- a/lib/LWP/Authen/Basic.pm
+++ b/lib/LWP/Authen/Basic.pm
@@ -8,6 +8,10 @@ sub auth_header {
     return "Basic " . MIME::Base64::encode("$user:$pass", "");
 }
 
+sub reauth_requested {
+    return 0;
+}
+
 sub authenticate
 {
     my($class, $ua, $proxy, $auth_param, $response,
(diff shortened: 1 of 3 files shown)


Frank071:
After pulling out several hairs I stumbled upon this ticket which describes the behaviour we are experiencing. What is blocking the commit of this patch? Can we help?

oalders:
@Frank071 are you able to test out the patch and let us know if it fixes this issue?

oalders:
I don't see that anyone has given this an "official" review.  Now that this issue has been poked, let's take a closer look at it and see what we can do.

Frank071:
@oalders I'll check both patches against the current CPAN version and get back to you

Frank071:
I checked the proposed patch. The patch for the two Authen:: modules could be applied unchanged, other than the linecount. The patch for the http.t test routine had to be rewritten, which I did. The patches can be found in the attached file. I am now running code that failed before (due to the expiry of the nonce) to check if it works in the real world as well. If so then I will try to turn the patch into proposed commits.

[digest_auth.patch.txt](https://github.com/libwww-perl/libwww-perl/files/3012675/digest_auth.patch.txt)

Frank071:
Patch checked in a working script and problem is solved. I split the pull by marmotil into two: #311 contains essentially the same patch as marmotil (so if partial acceptance of #40 is possible then this is the preferred way as credits go to marmotil). #312 contains an adapted version of the patch for http.t that marmotil suggested. In between 2012 and now http.t has been changed quite considerably so the original patch would not fit anymore. This one does ;-)

Suggestion to either approve #311 or (partial) #40 and #312 and to reject (and close) #94

oalders:
Thanks for testing this @Frank071 and also for the patches.  👍

oalders:
Closed via #313.  Thanks @marmotil!

PR #53, opened 2014-01-11
Check $ENV{TMPDIR} before other paths

dex4er:
On Android:

```
$ prove -v -bl t/local/get.t 
t/local/get.t .. 
1..0
ok 1
skipped: (no reason given)

Test Summary Report
-------------------
t/local/get.t (Wstat: 0 Tests: 1 Failed: 1)
  Failed test:  1
  Parse errors: Bad plan.  You planned 0 tests but ran 1.
```

It is because this test assumes that $TMPDIR is one of hardcoded dirs which don't exist on Android.

I suggest to use $TMPDIR before other dirs as workaround.

### DIFF
--- a/t/local/get.t
+++ b/t/local/get.t
@@ -10,7 +10,7 @@ if ($^O eq "MacOS") {
 
 # First locate some suitable tmp-dir.  We need an absolute path.
 $TMPDIR = undef;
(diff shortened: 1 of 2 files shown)


karenetheridge:
Merged ef85e61 as 83a5240b -- thanks!

PR #54, opened 2014-02-19
Synced up versions across files. Used package variables when available.

ironcamel:


oalders:
Sorry for the lack of feedback.  We're trying slowly to get caught up here.  The module version numbers are now in sync in master, but it would be handy to use the package versions for the bin scripts if you felt like reworking this pull request.

ironcamel:
No worries. I don't think it's a big deal. I forget now why I needed this.

PR #62, opened 2014-08-12
Fix for autoloading classes from various _proxy ENV vars

burak:
- Don't try to load an already tested module; this can trigger the
  "Attempt to reload %s aborted.\n" exception in newer perls.
- Handle the eval errors correctly by checking the eval return value, not just $@.

I've also changed the return statement position as $scheme sanity checks need to happen before interpreting its value.


simbabque:
Thank you for this @burak. I'm sorry it took so long, but your changes have just landed!

PR #65, opened 2014-12-12
recognize another error message from Socket

ghost:
when requesting a non-existing domain name, one would expect to see "Bad hostname", instead we only see the generic "Can't connect" because Socket returns the the error "nodename nor servname provided, or not known" in $@.


oalders:
LGTM

genio:
Thanks, @ghost

PR #66, opened 2015-01-09
change EAGAIN to EWOULDBLOCK to support Windows too

noxxi:
This is the same problem described in https://github.com/libwww-perl/net-http/pull/11 for Net::HTTP, but this time for libwww-perl. Both patches need to go together.

### DIFF
--- a/lib/LWP/Protocol/http.pm
+++ b/lib/LWP/Protocol/http.pm
@@ -282,7 +282,7 @@ sub request
             my $n = $socket->syswrite($req_buf, length($req_buf));
             unless (defined $n) {
                 redo WRITE if $!{EINTR};
-                if ($!{EAGAIN}) {
+                if ($!{EWOULDBLOCK}) {
                     select(undef, undef, undef, 0.1);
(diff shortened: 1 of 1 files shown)


karenetheridge:
+1

wchristian:
Any feedback on this? It's kind of important.

Zarabozo:
Wow. This issue was driving me crazy! I had to patch both Net/HTTP/Methods.pm and LWP/Protocol/HTTP.pm. The only workaround without patching them is to use IO::Socket::SSL version 2.005, which is the last version still not supporting non-blocking sockets on Windows.

karenetheridge:
I will do a release of libwww-perl at the end of the day if I don't hear
anything from Mike Schilli.

karenetheridge:
Merged; releasing imminently.

Zarabozo:
@karenetheridge Are you able to do the same for Net::HTTP?

karenetheridge:
No, I never received comaint of that when the libwww-perl transfer
happened, sorry...

PR #68, opened 2015-01-15
Document how to handle errors writing to the content file.

thrig:
Due to https://github.com/pfig/net-amazon-s3/issues/12 and how to check for the error being slightly non-intuitive. Only a POD patch; t/misc/pod.t tests still pass.


oalders:
Thanks @thrig!

PR #71, opened 2015-02-13
Test cleanup

haarg:
This cleans up the test running mechanism.  The t/TEST script was extra complexity that wasn't really needed, and communicating through flag files was annoying.

This allows the tests to be run from any directory and allows them to all be used without explicitly being enabled by flag files.  It then adjusts the Makefile.PL to pick the tests to run directly based on the arguments it was given, rather than communicating that to t/TEST via flag files.  Avoiding t/TEST also fixes the problems with nmake.

Fixes [RT#102083](https://rt.cpan.org/Public/Bug/Display.html?id=102083)


karenetheridge:
Awesome, thank you!

karenetheridge:
released in 6.11.

PR #72, opened 2015-03-13
Add HTML::Form dependency

dwmarshall:
Although HTML::Form isn't used for running the test suite (as was
the rationale for adding a dependency on HTTP::Daemon), we should add a
dependency on HTML::Form for the sake of backwards compatability (which
was the other rationale for the dependency on HTTP::Daemon).

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -48,6 +48,7 @@ my %WriteMakefileArgs = (
         'Encode::Locale' => 0,
         'File::Listing' => 6,
         'HTML::Entities' => 0,
+        'HTML::Form' => 6,
         'HTML::HeadParser' => 0,
         'HTTP::Cookies' => 6,
         'HTTP::Daemon' => 6,


karenetheridge:
Can you elaborate more on this backcompat issue?

dwmarshall:
LWP < 6 installed `HTML::Form`, as it did a lot of other modules. These were separated out for LWP 6. Gisle created a dependency on `HTTP::Daemon` for two reasons: first, because the LWP tests needed it and second, because code written for LWP < 6 that expected `HTML::Form` to be present might break. To whatever extent the second reason motivates adding a dependency on `HTTP::Daemon`, the same applies to `HTML::Form`.

This PR is possibly moot. Our specific problem was with `WWW::Mechanize`, but the problems we were having might be related to our own packaging system.

karenetheridge:
@gisle can you comment/review? I don't know enough about the history here to be sure this is a safe change.

gisle:
My preference would be to not add this dependency, simply because libwww-perl doesn't reference the module.  The change should be safe enough, but it prevents you from avoiding the module if you don't care for it.

dwmarshall:
I think Gisle's opinion is reasonable. Thank you for considering the change.

PR #73, opened 2015-05-14
Update timeouts on cached connections

kablamo:
Problem
-----------

If I do this
```
$ua->conn_cache($conn_cache);
$ua->timeout(1)
$ua->get($url);
# one connection is now cached.  the cached connection has a timeout of 1
$ua->timeout(6);
$ua->get($url); 
```

The second `get()` will timeout after 1 second when it should wait 6 seconds before timing out.

Solution
----------
My solution is to change `$ua->timeout` and `$ua->conn_cache` to loop through all the connections in `$conn_cache` and update the timeouts like this:
```
for my $conn ($conn_cache->get_connections) {
    $conn->timeout($new_timeout);
}
```

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -623,7 +623,21 @@ sub get_basic_credentials
 }
 
 
-sub timeout      { shift->_elem('timeout',      @_); }
+sub timeout
+{
+    my $self = shift;
+    my $old = $self->{timeout};
+    if (@_) {
+        $self->{timeout} = shift;
+        if (my $conn_cache = $self->conn_cache) {
+            for my $conn ($conn_cache->get_connections) {
+                $conn->timeout($self->{timeout});
+            }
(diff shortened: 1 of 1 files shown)


yuripats:
Any updates on this?

genio:
Hi @yuripats.  

The test suite is being updated and tested. Given that, I wouldn't expect this change to take place until the test suite in ```t/net/``` gets modernized and tests for this functionality are added.

```t/net/http-timeout.t``` will need to be updated to accommodate these changes once those current test files are updated and smoke tested successfully.

genio:
@yuripats Would it be possible for you to rebase this against master (removing the changes to Makefile.PL) and also to add a few tests now that the test suite is a bit more updated?

oalders:
I believe it's @kablamo who may want to rebase?

genio:
Yes. Sorry about that!

PR #78, opened 2015-09-02
FIX: continue challenge schemes

romangaleev:
When several challenge scheme is supported by web server (Negotiate and Basic for example)
continue trying next scheme if previous one was failed.

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -392,8 +392,10 @@ sub request
 				  "Unsupported authentication scheme '$scheme'");
 		next CHALLENGE;
 	    }
-	    return $class->authenticate($self, $proxy, $challenge, $response,
+	    my $re = $class->authenticate($self, $proxy, $challenge, $response,
 					$request, $arg, $size);
+		next CHALLENGE if ($re->code == 401);
+		return $re;
 	}
 	return $response;
     }


ioanrogers:
:+1:

simbabque:
I think this is a good idea. 

[The RFC](https://tools.ietf.org/html/rfc2617#section-4.6) is a bit unclear on whether we should keep going after we have supplied wrong credentials, but there are other schemes left. It does say to pick the strongest first, but we currently pick _Basic_ when both _Basic_ and _Digest_ are provided.

> An HTTP/1.1 server may return multiple challenges with a 401
   (Authenticate) response, and each challenge may use a different
   auth-scheme. A user agent MUST choose to use the strongest auth-
   scheme it understands and request credentials from the user based
   upon that challenge.

Given it is possible to implement all kinds of wickedness in `get_basic_credentials`, and that this method is used for (at least) _Digest_ too as far as I understand, I believe we cannot really pick the strongest type we understand.

What we are currently doing instead is go over all the ones the server provides in whatever order they are supplied in the `WWW-Authenticate` (or `Proxy-Authenticate`) header, and if we can use one, we do.

We do not keep going with the remaining schemes if the credentials are wrong.

This PR does not provide a test for this new behaviour. I've created a test case, and will make a new PR based on this one.

Issue #80, opened 2015-10-18
Remove LWP::Protocol::GHTTP


njtaylor:
HTTP::GHTTP this hasn't been updated since 2002, and uses libghttp from an old version of gnome, newer gnome versions have replaced libghttp with libsoup.

LWP::Protocol::GHTTP could be removed.

There are no tests to check this works.

genio:
An attempt at a separate module for LWP::Protocol::GHTTP has been created here https://github.com/genio/lwp-protocol-ghttp  If the idea of separating it this way isn't popular, the tests for it here: https://github.com/genio/lwp-protocol-ghttp/blob/master/t/00-load.t could be introduced into libwww-perl

genio:
The separate dist has been shipped so that a version of LWP::Protocol::GHTTP is out there and available to those who may still be using it.  So, step one of removing it from this dist has been completed.

genio:
GHTTP has been removed from this dist and will be gone in the next release!  Thanks everyone

PR #81, opened 2015-10-31
Remove LWP::Protocol::GHTTP

katkad:
because HTTP::GHTTP is obsoleted by obsolete libghttp

If you happen to remove LWP::Protocol::GHTTP according to https://github.com/libwww-perl/libwww-perl/issues/80 you can use this pull request.


karenetheridge:
I think this one is @gisle's call.

gisle:
I would not mind if it was unbundled from libwww-perl, but I then think `LWP::Protocol::GHTTP` ought to have its own dist on CPAN in case anybody actually uses it.  The `HTTP::GHTTP` module is still present.

genio:
```HTTP::GHTTP``` needs to be updated as it fails to install due to outdated locations for the library itself.

genio:
https://github.com/genio/lwp-protocol-ghttp

@gisle Can you review that and see what you think as a possible separate dist?

gisle:
Looks good to me.

genio:
@gisle Great!  If you want to release it so that it's easier to keep up with the rest of the libwww module set, please feel free to do so.  I have no strong feelings about doing so myself as all I have done is basically just package it up and add a test set.  On the other hand, if you prefer I do it, please let me know.

gisle:
I'm lazy — so I prefer if you do it. :-)

genio:
@karenetheridge I've scheduled my unauthorized push to the CPAN for deletion.  Sorry about being a bit click-happy and publishing without first thinking about PAUSE permissions.  Please let me know if you need anything from me in order to take over the github repo, etc.

genio:
[LWP::Protocol::GHTTP](https://metacpan.org/pod/LWP::Protocol::GHTTP) has been released and should be able to be safely removed from this dist now.

genio:
Thanks, @katkad

PR #85, opened 2015-12-05
Test::More with all tests in t/base and t/live

genio:
Make all tests in t/base use Test::More

Also, got rid of shebang lines for test files. Kind of formatted all three in a very similar way so they have the same feel to them when looking through them.  Ensured all tests have a description string.

### DIFF
--- a/t/base/protocols.t
+++ b/t/base/protocols.t
@@ -1,11 +1,11 @@
 use strict;
 use warnings;
 use Test::More;
-plan tests => 6;
 
 use LWP::UserAgent;
 
 my $ua = LWP::UserAgent->new();
+isa_ok($ua, 'LWP::UserAgent', 'new: UserAgent instance');
 
 $ua->protocols_forbidden(['hTtP']);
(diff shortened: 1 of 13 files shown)


t/base/protocols.t
```diff
@@ -18,3 +18,5 @@ ok(!$ua->protocols_allowed(),           '! $ua->protocols_allowed');
 
 $ua->protocols_forbidden(undef);
 ok(!$ua->protocols_forbidden(),         '$ua->protocols_forbidden(undef)');
+
+done_testing();
```
karenetheridge: done_testing requires Test::More 0.88, so either the prereq needs to be raised, or (more preferably) we stick with `plan`s.

genio:
I installed ```Test::More``` v0.80 and tested with the plans added back in with everything behaving as expected.  I'll be sure to remember that going forward.

genio:
Going to close this out and re-submit it later with proper branching and all of the tests updated.

PR #88, opened 2016-02-17
Test cleanup

genio:
Cleanup the entire test suite to use strict, warnings, and Test::More.  It seems to be testing well in 6.15_004

PR #90, opened 2016-03-03
Update "timeout" when reusing sockets

nfg:
Hiya!

This is a fix for https://rt.cpan.org/Public/Bug/Display.html?id=110893 . Basically, timeout is only set when the socket is created, so if a socket is reused, timeout wasn't updated.

I tested with this script: https://gist.github.com/nfg/2a8ef485b004524fe163

Cheers!

### DIFF
--- a/lib/LWP/Protocol/http.pm
+++ b/lib/LWP/Protocol/http.pm
@@ -181,6 +181,9 @@ sub request
 		$socket->close;
 		$socket = undef;
 	    } # else use $socket
+	    else {
+		$socket->timeout($timeout);
+	    }
 	}
     }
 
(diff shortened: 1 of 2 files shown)


nfg:
Right, I just noticed PR #73 . My fix is a little simpler, IMO. But it sounds like the blocker is test updates.

genio:
PR #88 addresses a lot of the testing cleanup and should hopefully make it easier to test things going forward if it's accepted.  Some or all of PR #88 may not be up-to-snuff and will likely take some time to be fully reviewed.

oalders:
@genio any reason not to merge this?

karenetheridge:
Looks good to me. Is it possible to write a test to demonstrate the issue/solution?

nfg:
Sure, I can take a stab at that. (Also, hey @karenetheridge!)

genio:
This looks good to me.

oalders:
Thanks @nfg!

PR #91, opened 2016-04-27
Encourage the use of HTTP::CookieJar::LWP

tomhukins:
HTTP::CookieJar::LWP supports https://publicsuffix.org/ whereas
HTTP::Cookies does not.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -16,6 +16,7 @@ on 'runtime' => sub {
     requires 'HTML::Entities';
     requires 'HTML::HeadParser';
     requires 'HTTP::Cookies' => '6';
+    requires 'HTTP::CookieJar::LWP';
     requires 'HTTP::Daemon' => '6';
     requires 'HTTP::Date' => '6';
     requires 'HTTP::Negotiate' => '6';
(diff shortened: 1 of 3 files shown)


oalders:
I wonder if it might be cleaner to pass a `cookie_jar_class` argument (which would default to `HTTP::Cookies`) to the constructor rather than relying on the user to remember the difference between an empty array ref vs empty hash ref?

oalders:
The arrayref vs hashref behaviour still makes a bit uneasy.  Is it cool if I push my proposed API change to this branch?  If we don't like it, I can delete the commit and we can move forward without it.

tomhukins:
On Thu, Apr 13, 2017 at 11:43:38AM -0700, Olaf Alders wrote:
> The arrayref vs hashref behaviour still makes a bit uneasy.  Is it
> cool if I push my proposed API change to this branch?  If we don't
> like it, I can delete the commit and we can move forward without it.

Please do.  Several people have told me they dislike the
arrayref/hashref interface.  I experimented with your idea but couldn't
find an implementation that I liked in the time I spent.  I look forward
to seeing what you come up with.

simbabque:
I realise it has been a while, but this is a great idea and we should have something to this accord, where we allow the user to default to the best practice without breaking backwards compatibility. The documentation was already updated by Olaf a few years after you submitted this PR, so there's a conflict now. I don't know if Olaf ever got around to writing down what his idea for an API would look like, but he's sitting next to me at the PTS so I will poke him to check.

simbabque:
Thank you @tomhukins!

PR #92, opened 2016-05-04
Allow LWP::UserAgent to send headers on CONNECT requests.

ghost:
This pull commit includes changes for LWP::UserAgent and LWP::Protocol::http, respectively. It adds functionality to specify headers to be sent to the proxy on a CONNECT request.

This corresponds to curl's new "--proxy-header" option.

Usage:

```
require LWP::UserAgent;

my $ua = LWP::UserAgent->new;
$ua->proxy(['http', 'https'], 'http://proxy.example.com');
$ua->proxy_header('User-Agent' => 'Special Agent');
```


voldemortensen:
LGTM

mafopen:
Thanks for confirming.
Does anyone else have any objections?

oalders:
@mafopen could we get an accompanying test for this?

lib/LWP/UserAgent.pm
```diff
@@ -712,13 +712,19 @@ sub cookie_jar {
...
 }
 
-sub get_proxy_headers {
+sub proxy_headers {
     my $self = shift;
-    return $self->{proxy_headers};
+    my $old = $self->{proxy_headers} ||= HTTP::Headers->new;
+    if (@_) {
+	Carp::croak("proxy_headers not set to HTTP::Headers compatible object")
+	    unless @_ == 1 && $_[0]->can("header_field_names");
```
oalders: Should we check that this is an object before we call `can` on it?

oalders (dismissed):

lib/LWP/UserAgent.pm
```diff
@@ -747,6 +747,23 @@ sub cookie_jar {
...
+    $self->{proxy_headers} ||= HTTP::Headers->new;
+    return $self->{proxy_headers}->header(@_);
+}
+
+sub proxy_headers {
+    my $self = shift;
+    my $old = $self->{proxy_headers} ||= HTTP::Headers->new;
+    if (@_) {
+	Carp::croak("proxy_headers not set to HTTP::Headers compatible object")
+	    unless @_ == 1 && $_[0]->can("header_field_names");
```
oalders: Should we check that this is an object before we call `can` on it?

oalders (changes requested):

lib/LWP/UserAgent.pm
mafopen: I have use the same checks as `default_headers` uses. Usage of `proxy_headers` and `default_headers` should be the same, shouldn't it?

oalders: Yes, I'm just not sure the original is correct in this case.  `->can()` assumes that an object has been passed.  If something else is passed to this method then we'll lose the intended error message.  Something like `unless @_ == 1 && blessed( $_[0] ) && $_[0]->can('header_field_names')` would cover that case.

oalders:
@genio just pointed out the following:

```
[09:37:29]  <genio>	Shouldn't proxy_headers be allowed to be passed in the constructor as well (a la default_headers)?
```

For the sake of consistency, I think it would make sense to allow `proxy_headers` to be passed to `new()`.

mafopen:
@genio & @oalders , I'm happy to add `proxy_headers` to the constructor as well, if you agree.
What's your expectation of what options should be available through the constructor. This is probably a rather rare case and it may be confusing to list it. There's other options like `no_proxy` that aren't available either. But as I said, happy to add it if you like, I just didn't wanna be too intrusive with my changes.

mafopen:
@genio & @oalders , please let me know if I should add `proxy_headers` to the constructor. As pointed out above, there are many options that are not passed to the constructor and `proxy_header` is most likely one of the more rare cases.
If you think I should add it I will do so of course. If not, could you please go ahead and merge this PR?

genio:
@mafopen Sorry, this one completely slipped through the cracks. Re-reading, I think you're right and given that this is a rare scenario, I'm ok with it not being available in the constructor.  That being said, would you mind writing a bit of documentation for the extra methods?

Once documentation is done, I wouldn't have a problem merging this.

Sorry again for letting this fall through.

karenetheridge:
(obligatory: when everything is done, please squash the commits and force-push.)

oalders:
@mafopen I'm sorry that I let this one slip.  If you're able to get these changes in, I'll make it a priority to merge and release.

mafopen:
What documentation are you after? I have updated the pod documentation as well.

PR #100, opened 2016-12-31
Let default_header(Content-Type => '...') work for post/put

wolfsage:
If you say:

    $ua->default_header('Content-Type' => 'foo');
    $ua->post($url, Content => 'string');

You very likely mean for the request to use the provided
default Content-Type (in this case: 'foo').

This change allows that to happen, while still letting the
following forms override the default Content-Type with
'x-www-form-urlencoded' (since that makes sense as the body
of the request is being encoded for you):

    $ua->post($url, \%form);
    $ua->post($url, \@form);
    $ua->post($url, Content => \%form);
    $ua->post($url, Content => \@form);

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -411,12 +411,47 @@ sub get {
     return $self->request( HTTP::Request::Common::GET( @parameters ), @suff );
 }
 
+sub _has_raw_content {
+    my $self = shift;
+    shift; # drop url
+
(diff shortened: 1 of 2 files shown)


wolfsage:
For https://rt.cpan.org/Public/Bug/Display.html?id=111276

oalders:
👍

oalders:
Thanks @wolfsage!

PR #105, opened 2017-01-27
Eval -> Try::Tiny

genio:
Update all of the places where we had a raw ```eval``` and replace it with a ```try {} catch {};``` block.

There was also a ```UNIVERSAL::isa()``` call that got switched to ```->isa()``` instead.  These changes require adding two prerequisites, ```Scalar::Util``` and ```Try::Tiny```.

While the changes test properly, they did require a few changes to the loading of protocol packages.

A full review by a few is needed.


oalders:
LGTM

Issue #107, opened 2017-02-03
You can't call return from within a Try::Tiny catch block


skington:
As a result, calling LWP::UserAgent->simple_request with a malformed URL does *not* return a faked "that was a malformed URL" response, but proceeds to call $self->send_request instead.

I believe the problem was introduced [when eval was replaced by try/catch](https://github.com/libwww-perl/libwww-perl/commit/650853e6948aafad299d64714cfcaae169e9cf27#diff-e316f49c673f6e1c852d5b23a10e4b69R270). The problem is that [Try::Tiny does not support returning from the parent subroutine](https://github.com/p5sagit/Try-Tiny/blob/d8d64fc24d7b863f9b887d9e6eceacf52f2f9ea4/lib/Try/Tiny.pm#L229-L231) so the return statement is ignored.

karenetheridge:
I'm a bit surprised we don't have a test for that error.. we should fix
that too :)

genio:
@skington Oooh. We'll create a test and a fix as quickly as possible.  Thanks for pointing this out

genio:
I think https://github.com/libwww-perl/libwww-perl/compare/genio/fixtry should accurately test for and fix the problem.

@skington Can you try that patch and ensure it works out the issue in your usage, please?

skington:
Won't that also return a malformed request response if there's an error when sending, though?

genio:
I think you're right.  Sorry about that, I think I'm moving too fast and not thinking enough. https://github.com/libwww-perl/libwww-perl/compare/genio/fixtry should be better now.

skington:
That works for me, yes.

oalders:
Closed via #108

PR #108, opened 2017-02-03
Fix incorrect usage of Try::Tiny

genio:
As was pointed out in Issue #107 an error was introduced with moving from raw ```eval```s to ```Try::Tiny``` within ```LWP::UserAgent->simple_request```.

This fix should accurately trap the errors and return the error response when found.  Also, a few new tests were added to ```t/base/ua.t``` to test the ways in which the call to ```->prepare_request``` can die from within ```->simple_request```.

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -267,14 +267,18 @@ sub simple_request
         Carp::croak("No request object passed in");
     }
 
+    my $error;
     try {
         $request = $self->prepare_request($request);
     }
     catch {
-        my $error = $_;
+        $error = $_;
         $error =~ s/ at .* line \d+.*//s;  # remove file/line number
-        return _new_response($request, &HTTP::Status::RC_BAD_REQUEST, $error);
     };
+
+    if ($error) {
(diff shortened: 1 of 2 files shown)


skington:
There's [at least one other similar bug](https://github.com/libwww-perl/libwww-perl/blob/2c94749a8f9b9f7bc2c2cb74169b213381f859b5/lib/LWP/UserAgent.pm#L408-L409).

genio:
@skington I'll move the ```next``` call outside of the ```catch``` block.

skington:
That's not the problem I was referring to; rather, you have a return statement inside a catch block, which won't actually return from `sub request`, so instead it will return $response instead, which was not what was intended.

genio:
@skington Ah. I can see where you'd think that.  However, the mix of tabs and spaces are throwing off where you're seeing the end of the ```catch {}``` block.  That block ends here: https://github.com/libwww-perl/libwww-perl/blob/2c94749a8f9b9f7bc2c2cb74169b213381f859b5/lib/LWP/UserAgent.pm#L401

PR #120, opened 2017-02-28
Doc differences for HTTP::Message < 6.07

genio:
Update the documentation to make it clear that certain uses of the ```->put()``` method require at least HTTP::Message 6.07.

Also, clean up the documentation a bit to make it easier to find the methods in their ```=head2``` directive rather than multiple ```=item``` lines.


oalders:
This was a monster.  Thank you @genio!

PR #125, opened 2017-03-13
Allow setting of proxy and no_proxy from the constructor of LWP::UA

### LINKED ISSUE #124: Proxy attributes should be settable from the constructor
Almost all the other attributes are settable from the constructor. I cannot think of a good reason why `proxy`, `no_proxy`, and `env_proxy` should be different in this regard.

Less preferably, the documentation could be improved to explicitly mention that these are not settable from the constructor.

oschwald:
Closes #124

### DIFF
--- a/t/base/ua.t
+++ b/t/base/ua.t
@@ -15,6 +15,8 @@ delete $ENV{PERL_LWP_SSL_CA_PATH};
 delete $ENV{PERL_LWP_ENV_PROXY};
 
 subtest 'proxy settings from the constructor' => sub {
+    plan tests => 4;
+
     my $ua = LWP::UserAgent->new(
         proxy => [
             ftp => 'http://www.sol.no',
--- a/t/base/ua.t
+++ b/t/base/ua.t
@@ -4,7 +4,7 @@ use HTTP::Request ();
 use LWP::UserAgent ();
 use Test::More;
 
-plan tests => 40;
+plan tests => 41;
 
 # Prevent environment from interfering with test:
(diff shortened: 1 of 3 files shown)


oalders:
This looks good to me.  My only concern would be whether we want to introduce a dependency on `List::Util`.  It doesn't do so well on Perl 5.8 in addition to some FreeBSD failures with more modern Perls.  http://matrix.cpantesters.org/?dist=Scalar-List-Utils+1.47

oschwald:
`List::Util` has been part of core since 5.8, although I am not sure if `pairmap` goes back that far.

oschwald:
It looks like it was [added in 2013](https://metacpan.org/changes/distribution/Scalar-List-Utils#L155) so it isn't in 5.8. 5.8.9 has 1.19.

oschwald:
I pushed a version that does not use `pairmap`. Feel free to squash on merge.

oschwald:
I added the subtest plan in fbb5a5a.

t/base/ua.t
```diff
@@ -14,6 +14,25 @@ delete $ENV{PERL_LWP_SSL_CA_FILE};
 delete $ENV{PERL_LWP_SSL_CA_PATH};
 delete $ENV{PERL_LWP_ENV_PROXY};
 
+subtest 'proxy settings from the constructor' => sub {
+    plan tests => 4;
```
genio: That should be 3, right?

oschwald: No, see the postfix for.

genio: gah. right. sorry.

oschwald:
Thanks for merging!

PR #126, opened 2017-03-29
add send_te configuration to disable sending the TE header

### LINKED ISSUE #27: Wishlist: option to turn off sending TE header
Since a few weeks ago, my internet banking module Finance::Bank::ID::BCA fails with Connection Reset error when it tries to connect to https://ibank.klikbca.com/ . The module uses WWW::Mechanize, which in turns uses LWP::UserAgent.

A closer look reveals that the server is now rejecting TE request header. It would be nice if LWP::Protocol::http lets us customize whether to send the header or not.

FYI, the server also rejects HTTP/1.0 protocol.

Thanks.

Ref: http://www.perlmonks.org/?node_id=957926

preaction:
Fixes #27

### DIFF
--- a/lib/LWP/Protocol/http.pm
+++ b/lib/LWP/Protocol/http.pm
@@ -30,7 +30,7 @@ sub _new_socket
 					Proto    => 'tcp',
 					Timeout  => $timeout,
 					KeepAlive => !!$self->{ua}{conn_cache},
-					SendTE    => 1,
+					SendTE    => $self->{ua}->send_te,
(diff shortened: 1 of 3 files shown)


genio:
The changes look fine to me. We'll need to document this extra attribute and how it ties into ```::Protocol::http```.

oalders:
@preaction do you have time to document the changes or would you like one of us to do this?

preaction:
I've added docs and rebased

oalders:
Thanks @preaction!

Issue #86, opened 2015-12-23
LWP::UserAgent does not check Content-Length header


skaji:
Even if Content-Length header is not equal to the actual content length,
LWP::UserAgent::request() returns a 200 OK response.

Here is an example:

``` perl
# test.pl
use strict;
use warnings;
use LWP::UserAgent;
use Plack::Loader;
use Test::TCP;

my $server = Test::TCP->new(code => sub {
    my $port = shift;
    my $app = sub {
        my $content = "hello world\n";
        [
            200,
            ['Content-Length' => 10000], # bigger than acutual content length
            [$content],
        ];
    };
    Plack::Loader->load('Standalone', port => $port)->run($app);
});

my $url = "http://localhost:" . $server->port;
my $http = LWP::UserAgent->new;
my $res = $http->get($url);
warn $res->status_line, "\n";
```

```
> perl test.pl
200 OK
```

As an end-user, I want LWP::UserAgent to check Content-Length header and if it is not equal to the actual content length, I want LWP::UserAgent to return a response which indicates failure.

damog:
Can you please refer to the HTTP spec section where the client needs to err given the mismatch on body length and header?

cavac:
A wrong content length interferes with connection reuse an may be a potential security problem.

For example, a server script may generate a content that also includes headers+body. By manipulating the content length header to make the number smaller (which the server shouldn't allow anyway), the server script may hijack the next client request over the same TCP connection by providing fake headers plus a potentially malicious content.

FGasper:
Confirmed still happening in 6.24.

genio:
We've looked into this several ways now. We also see the same behavior with Mojo. There doesn't appear to be a clean way to fix it. Every simple approach leaves things a bit sketchy and in some ways, outright broken. @skaji you mind taking a stab at it?

oalders:
@skaji just wondering why this was closed?

skaji:
I closed all issues and pull requests created by me that I am not currently intending to address.

If you think this issue should be left open, please feel free to reopen it.

PR #210, opened 2017-03-31
warn() when Content-Length exceeds the actual payload length.

FGasper:
Issue #86: This matches the behavior described in the last
paragraph of RFC 2616, section 4.4.

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -422,6 +422,12 @@ sub request {
         }
         return $response;
     }
+
+    my $content_length = $response->header('content-length');
+    if (defined $content_length && (length($response->{'_content'}) > $content_length)) {
+        warn sprintf( "'Content-Length' header (%d) exceeds the delivered payload length (%d)!", $content_length, length($response->{'_content'}) );
+    }
+
     return $response;
 }
 


skaji:
@FGasper thanks for this.
* Not ">" but "<"
* Can we use "Client-Warning" header for this?
* Can we safely use a private variable "_content" here?

FGasper:
@skaji Why “<”? The point is to tell the caller (i.e., the client) that the server said they’d send 10,000 bytes but only gave 20 (or 9,998, or whatever). “>” is the correct operator for that.

The condition where there’s more data sent than advertised is a similar case that I filed against Net::HTTP.

I don’t know what the “Client-Warning” header is … HTTP 1.1 doesn’t define it, AFAICT.

_content is used in the same block where the diff adds code … it seems safe to me.

genio:
Could you add an appropriate test in maybe ```t/local/http.t```?

FGasper:
@genio I’ll see about finding time for it. When I looked through the existing tests before I didn’t see any patterns that suggested a good way to mock this behavior.

skaji:
@FGasper 
> Why “<”? The point is to tell the caller (i.e., the client) that the server said they’d send 10,000 bytes but only gave 20 (or 9,998, or whatever). “>” is the correct operator for that.
The condition where there’s more data sent than advertised is a similar case that I filed against Net::HTTP.

FYI: It seems that `length($response->content) > $response->header('Content-Length')` does not occur in current LWP (or Net::HTTP) implementation.

> I don’t know what the “Client-Warning” header is … HTTP 1.1 doesn’t define it, AFAICT.

Client-Warning is a pseudo HTTP header by LWP. 
eg: https://github.com/libwww-perl/libwww-perl/blob/master/lib/LWP/UserAgent.pm#L301

> _content is used in the same block where the diff adds code … it seems safe to me.

`$response->{_content}` may be a scalar reference.
https://github.com/libwww-perl/HTTP-Message/blob/master/lib/HTTP/Message.pm
I would use a documented public method `$response->content` or `$response->content_ref`.

FGasper:
@skaji `length($response->content) > $response->header('Content-Length')` happens every time a payload like what the test example in the OP happens.

skaji:
@FGasper Sorry, what is `the test example in the OP`?

FGasper:
@skaji https://github.com/libwww-perl/libwww-perl/issues/86

skaji:
Do you mean this?
```perl
# test.pl
use strict;
use warnings;
use LWP::UserAgent;
use Plack::Loader;
use Test::TCP;

my $server = Test::TCP->new(code => sub {
    my $port = shift;
    my $app = sub {
        my $content = "hello world\n";
        [
            200,
            ['Content-Length' => 10000], # bigger than acutual content length
            [$content],
        ];
    };
    Plack::Loader->load('Standalone', port => $port)->run($app);
});

my $url = "http://localhost:" . $server->port;
my $http = LWP::UserAgent->new;
my $res = $http->get($url);
warn $res->status_line, "\n";

warn q|length($res->content) = |, length($res->content), "\n";
warn q|$res->header('Content-Length') = |, $res->header('Content-Length'), "\n";
```

Then:
```
❯ perl test.pl
200 OK
length($res->content) = 12
$res->header('Content-Length') = 10000
```
which is `length($res->content) < $res->header('Content-Length')`.

FGasper:
@skaji Ah, you’re right … I must have unwittingly changed the diff from what did fix the issue when I observed it.

I’ll resubmit later.

genio:
After looking at many things with another popular Perl user agent, testing the length like this isn't the way to go as it could introduce other issues. GZipped content, etc.  I think I'm going to close out this attempt for now and revisit it a bit in the IRC channel to hash out other ideas, but I don't see a way we can use this implementation.

Thank you for the attempt and please don't hesitate to try other methods.

-- Chase

Issue #184, opened 2017-03-31
bug of lwp-request(Use of uninitialized value $vals) [rt.cpan.org #92633]


oalders:
Migrated from [rt.cpan.org#92633](https://rt.cpan.org/Ticket/Display.html?id=92633) (status was 'new')

Requestors:
* dynkin@gmail.com

From dynkin@gmail.com on 2014-02-01 03:59:24:
```
I'm using following version of lwp-request

$ GET -v
This is lwp-request version 6.03 (libwww-perl-6.03)

Copyright 1995-1999, Gisle Aas.

This program is free software; you can redistribute it and/or
modify it under the same terms as Perl itself.


When I GET the following url with '-ed' option, I get two warnings
related to uninitialized values.


$ GET -ed http://www.bbc.co.uk/news/business-25956284 | head
Use of uninitialized value $vals in index at
/usr/share/perl5/HTTP/Headers.pm line 259.
Use of uninitialized value $vals in concatenation (.) or string at
/usr/share/perl5/HTTP/Headers.pm line 262.
200 OK
Cache-Control: private, max-age=0, must-revalidate
Connection: close
...
[snip]



thanks!

-- 
george
```

genio:
This will be corrected in the next release via https://github.com/libwww-perl/libwww-perl/pull/249

PR #249, opened 2017-04-06
use strict and warnings in all bin files and perltidy them

genio:
This would address issue #184 

* Gets rid of the ```-w``` flag.
* Adds ```use strict;```
* Adds ```use warnigns;```
* Defines all variables where necessary,
* Cleans up code with perltidy.

### DIFF
--- a/bin/lwp-request
+++ b/bin/lwp-request
@@ -181,8 +181,8 @@ Gisle Aas <gisle@aas.no>
 =cut
 
 $progname = $0;
-$progname =~ s,.*[\\/],,;  # use basename only
-$progname =~ s/\.\w*$//;   # strip extension, if any
+$progname =~ s,.*[\\/],,;    # use basename only
+$progname =~ s/\.\w*$//;     # strip extension, if any
 
 require LWP;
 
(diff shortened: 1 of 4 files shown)


petdance:
👍  on the first four bullets but  👎 on the tidy part.  That seems like a lot of diff churn.

If it's worth doing the tidying, I'd suggest doing it as a separate patch that is just tidy changes.

oalders:
For a big tidy it's easier to understand if the tidy happens in its own commit.  I'm not sure that we need to roll this back, but moving forward it would be a good thing to keep in mind.

petdance:
>  I'm not sure that we need to roll this back

It hasn't been committed yet, so we should still be able to have this as two separate commits.

genio:
I reverted the last attempt and split the effort into two separate commits to make the actual changes easier to view.

bin/lwp-request
```diff
@@ -364,8 +370,8 @@ if (defined $options{'i'}) {
     $options{'i'} = time2str($time);
 }
 
-$content = undef;
-$user_ct = undef;
+my $content = undef;
+my $user_ct = undef;
```
karenetheridge: `= undef` is redundant and a bit silly; let's remove that.

karenetheridge:
I lean slightly towards the "not doing it" side when it comes to tidying -- the churn is great, which makes it harder to perform future archaeology expeditions with `git blame` when debugging things.

Also, FWIW, I'm against vertically-aligning `=` and `=>` operators, as it inevitably results in more churn later when a single line is changed then the programmer feels compelled to re-align all the other lines again to match up.  I would stick with only adjusting leading whitespace to use a consistent four character indentation, and leave it at that.  Running code through `perltidy` or another automatic mangler always changes too much IMO.

oalders:
The first tidy is always the worst.  After that, it gets a lot better.  :)  I admit it has its problems, but my feeling is that (for me) it brings more positives than negatives.

We have to be judicious, though, since tidying wholesale can create a bunch of merge conflicts in outstanding pull requests.  In this case that won't happen, but it's something to keep in mind.

petdance:
It's not about lining up the `=` and `=>` so much as it is putting things to read into two distinct columns.  The value increases dramatically as the number of lines increases.  Two or three, no biggie.  Many lines of hash constructor or variable assignments?  It's night & day, and I'm glad to take the minimal cost of realigning.  

```
my %mech_parms = (
    autocheck => ($class eq 'WWW::Mechanize' ? 1 : 0),
    onwarn => \&WWW::Mechanize::_warn,
    onerror => \&WWW::Mechanize::_die,
    quiet => 0,
    stack_depth => 8675309,     # Arbitrarily humongous stack
    headers => {},
    noproxy => 0,
);

my %mech_parms = (
    autocheck   => ($class eq 'WWW::Mechanize' ? 1 : 0),
    onwarn      => \&WWW::Mechanize::_warn,
    onerror     => \&WWW::Mechanize::_die,
    quiet       => 0,
    stack_depth => 8675309,     # Arbitrarily humongous stack
    headers     => {},
    noproxy     => 0,
);
```

Remember also that `git diff` has many options for ignoring whitespace changes.  What made me concerned about this tidying on this PR was that some of the tidying had lines getting split between lines.

Issue #157, opened 2017-03-31
credentials method needs undef check on lc()


oalders:
Migrated from [rt.cpan.org#69339](https://rt.cpan.org/Ticket/Display.html?id=69339) (status was 'new')

Requestors:
* paul.a.liebert@hp.com

From paul.a.liebert@hp.com on 2011-07-07 15:28:21:
```
Use of uninitialized value within @_ in lc at LWP/UserAgent.pm line 581.

Starting with Perl 5.12, the builtin lc() is now more strict about
checking for undef arguments.  This particular bug was triggered by
the test suite when building WWW::Mechanize 1.62

Distro: libwww-perl-6.02.tar.gz

Perl:   This is perl 5, version 12, subversion 1 (v5.12.1)
        built for x86_64-linuxperl-5.12.1

OS:     Linux lxcviby.cv.hp.com 2.6.9-78.0.17.ELsmp #1 SMP
        Thu Mar 5 04:58:34 EST 2009 x86_64 x86_64 x86_64 GNU/Linux
```

simbabque:
This ticket is done and can be closed.

PR #251, opened 2017-04-13
lc() on possibly undefined value in ->credentials

genio:
This PR should help with: https://github.com/libwww-perl/libwww-perl/issues/157

In ```->credentials``` in ```LWP::UserAgent``` we do ```my $netloc = lc(shift);```. This value could be undefined and when it's so, it causes warnings.

This is a proposal for a fix for that, ```my $netloc = lc(shift || '');``` but I can see going with other options as well.  Please review this and supply your thoughts.

Also, I started another test file ```t/10-attrs.t``` to allow us to start exercising every attribute to hopefully get the coverage up a bit and to help us when fixing documentation later.

-- Chase

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,9 @@
 Change history for libwww-perl
 
 {{$NEXT}}
+    - Handle undefined values in ->credentials (GH #157)
+
+6.26      2017-04-12
     - Perltidy all apps in the bin/ directory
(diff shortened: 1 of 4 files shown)


genio:
I'll work up another PR to re-document that method with better examples after a direction is chosen to go about cleaning up this error.

PR #252, opened 2017-04-19
Add tests for the "get" & "head" functions

JRaspass:
This commits adds minimal tests for the get & head functional interface.
It extends the file that was already testing getprint & getstore.


oalders:
Thanks @JRaspass!

PR #253, opened 2017-04-19
Test LWP has a Version sub/method

JRaspass:
A simple test to make sure LWP->Version or LWP::Version() returns
the same as $LWP::VERSION.

### DIFF
--- a/t/base/version.t
+++ b/t/base/version.t
@@ -1,0 +1,8 @@
+use strict;
+use warnings;
+use LWP;
+use Test::More;
+
+plan tests => 1;
+
+is(LWP::Version(), $LWP::VERSION, 'LWP::Version() returns the same as $LWP::VERSION');


karenetheridge:
This `Version` sub is undocumented -- can we just get rid of it?

Its use is very limited:

http://grep.cpan.me/?q=LWP%3A%3AVersion
http://grep.cpan.me/?q=LWP-%3EVersion

genio:
I would love to get rid of it.  I'll see about putting in PRs to the few things making use of it later today.

JRaspass:
Dropping any uses in favour of `$LWP::VERSION`, `LWP::VERSION`, or `LWP->VERSION` seems sane, since the coderef always exists and just returns the value of the package var.

karenetheridge:
The `VERSION()` sub from UNIVERSAL does more than that -- when passed a value, it asserts that the current version satisfies the argument.

    LWP->VERSION('1.0');  # returns true
    LWP->VERSION('999');  # dies

This is commonly seen in code that conditionally requires a module for a feature:

    if (eval { require Foo; Foo->VERSION('1.23'); }) {
        ... # use the feature that was added in Foo 1.23...
    }

genio:
Our own bin scripts have been fixed.
1. ```lwp-mirror```
1. ```lwp-request```

Other dists doing the wrong thing...
1. [RT# 121280 for Crypt::NSS](https://rt.cpan.org/Ticket/Display.html?id=121280)
1. [RT# 121281 for CPAN::SQLite](https://rt.cpan.org/Ticket/Display.html?id=121281)
1. [RT# 121282 for WebFS::FileCopy](https://rt.cpan.org/Ticket/Display.html?id=121282)

JRaspass:
Re-running the greps since grep.cpan.me no longer exists:

 2 hits - https://grep.metacpan.org/search?q=LWP::Version
 1 hit - https://grep.metacpan.org/search?q=LWP->Version
 
Looks like the [CPAN::SQLite](https://rt.cpan.org/Ticket/Display.html?id=121281) issue is a false positive and the `WebFS::FileCopy` use can't work as there's no package called that anymore so we're really only down to one upstream use of `LWP->Version`. That use is also only in a test so I'm tempted to rewrite this PR to just drop the `Version` sub, it's been 5 years, what do people think?

oalders:
I think we can get rid of the sub. Anyone opposed to this?

skaji:
+1 for getting rid of the sub.

JRaspass:
Turns out I can't update this PR as it was forked under an organisation I no longer belong to so I'm closing this in favour of #416

Issue #130, opened 2017-03-30
Error in Authentication Digest Implementation [rt.cpan.org #20889]


oalders:
Migrated from [rt.cpan.org#20889](https://rt.cpan.org/Ticket/Display.html?id=20889) (status was 'new')

Requestors:
* perl@intertivityNOSP4M.com

From perl@intertivitynosp4m.com on 2006-08-08 13:17:57:
```
Hi,

RFC 2617 says to calculate the digest of the body only if the auth-int 
is set in the qop option. The current version of LWP::Authen::Digest 
just ignores this and caculates the digest always when the method is 
POST or PUT.

The following lines will fix it:

my %resp = map { $_ => $auth_param->{$_} } qw(realm nonce opaque);
    @resp{qw(username uri response algorithm)} = ($user, $uri, 
$digest, "MD5");

# changes start here

	my $auth_qop = $auth_param->{qop} || "";
    if ($auth_qop eq "auth") {
	@resp{qw(qop cnonce nc)} = ("auth", $cnonce, $nc);
    }

    my(@order) = qw(username realm qop algorithm uri nonce nc cnonce 
response);
    if($request->method =~ /^(?:POST|PUT)$/ && $auth_qop eq 'auth-
int' ) {

# changes end here

	$md5->add($request->content);
	my $content = $md5->hexdigest;


HTH
esskar
```

PR #254, opened 2017-04-20
Calculate digest of body only if auth-int specified.

colinnewell:
This applies the roughly the patch from issue #130 and adds tests to
check the changes work correctly.

This isn't quite complete yet.  I still want to do the following,

* Test against a live web server manually to verify the original issue and that servers definitely work with this behaviour.
* Consider adding a mechanism to allow for the legacy mode of operation in case there are also legitimate servers in the wild that do work in the way LWP was operating.

### DIFF
--- a/lib/LWP/Authen/Digest.pm
+++ b/lib/LWP/Authen/Digest.pm
@@ -26,7 +26,7 @@ sub auth_header {
     push(@digest, $auth_param->{nonce});
 
     if ($auth_param->{qop}) {
-	push(@digest, $nc, $cnonce, ($auth_param->{qop} =~ m|^auth[,;]auth-int$|) ? 'auth' : $auth_param->{qop});
+	push(@digest, $nc, $cnonce, ($auth_param->{qop} =~ m|^auth[,;]\s*auth-int$|) ? 'auth' : $auth_param->{qop});
     }
 
     $md5->add(join(":", $request->method, $uri));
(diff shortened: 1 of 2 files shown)


colinnewell:
I'm actually struggling to find a web server that supports auth-int.  I'm assuming they are out there because people complain about issues with authenticating with them, but the big 3 don't appear to support it.  Apache, nginx and IIS don't appear to support that option.

genio:
Auth-int with apache requires http://httpd.apache.org/docs/2.2/mod/mod_auth_digest.xml  I'm not aware of a public and available test server for digest auth though.

colinnewell:
Note the footnote on that though.  I got an apache server going and was talking to it before I came across that in the fine detail.

> auth-int is not implemented yet.

genio:
Ah. You're right. I didn't even notice that.

genio:
Maybe LightHTTP  http://redmine.lighttpd.net/issues/2478

colinnewell:
Again, doesn't look like that has support.  I've only looked at the commit related to the bug fix, but in that it blocks it off with a 'not supported' message.

> Just use https if you want to protect the connection; qop=auth-int won't be supported.

colinnewell:
It sort of looks like WEBrick might support it.

https://searchcode.com/file/12782721/lib/ruby/1.8/webrick/httpauth/digestauth.rb

genio:
I'm starting to wonder if we should drop support for it since it seems to be a non-real-world feature.

colinnewell:
I'm half with you on that, but I'll give it a bit more of a go before abandoning this as a bad job.  I could see complaints with other client libraries reporting the same problem so this isn't a completely isolated report, it's just not a very common option.

simbabque:
Did you ever look at this again @colinnewell? I'm working through old PRs right now to clean up our queue. Also, hi :wave:

colinnewell:
Sorry, I've run out of tuits to work on this.  Please do feel free to do whatever you need to with this.

Issue #201, opened 2017-03-31
Handling failed authentication when multiple auth methods are possible [rt.cpan.org #112465]


oalders:
Migrated from [rt.cpan.org#112465](https://rt.cpan.org/Ticket/Display.html?id=112465) (status was 'new')

Requestors:
* Slava.Goltser@ipsoft.com

From slava.goltser@ipsoft.com on 2016-02-26 22:09:07:
```
Below patch addresses the issue present in LWP::UserAgent 6.15 where if LWP::UserAgent::request successfully loads the first LWP::Authen submodule, all others are ignored, even if the first module fails.


The specific situation is when a server (in this case IIS) returns the following header:

  HTTP/1.1 401 Unauthorized
  Content-Type: text/html
  Server: Microsoft-IIS/7.5
  WWW-Authenticate: Negotiate
  WWW-Authenticate: NTLM
  X-Powered-By: ASP.NET

If both LWP::Authen::Negotiate and LWP::Authen::NTLM are present, LWP::UserAgent will try to load the first one and once (in this case) LWP::Authen::Negotiate fails, LWP::UserAgent returns that response without trying the other authentication modules/methods. In this case, the credentials were given for NTLM authentication which do work.


--- a/LWP/UserAgent.pm 2015-12-05 05:57:39.000000000 +0000
+++ b/lib/LWP/UserAgent.pm 2016-02-26 19:16:55.893851729 +0000
@@ -392,8 +392,13 @@
   "Unsupported authentication scheme '$scheme'");
  next CHALLENGE;
     }
-    return $class->authenticate($self, $proxy, $challenge, $response,
+    $response = $class->authenticate($self, $proxy, $challenge, $response,
  $request, $arg, $size);
+    # Try next challenge type if current failed
+    if (!$response->is_success) {
+ next CHALLENGE;
+    }
+    return $response;
  }
  return $response;
     }


--
Slava Goltser
IPsoft, Inc.
slava.goltser@ipsoft.com
```

Issue #248, opened 2017-03-31
Fwd: LWP small fix required [rt.cpan.org #120047]


oalders:
Migrated from [rt.cpan.org#120047](https://rt.cpan.org/Ticket/Display.html?id=120047) (status was 'open')

Requestors:
* Ladygin@srt-web.com

From ladygin@srt-web.com on 2017-01-31 08:43:46:
```
LWP Digest authorisation small bag, in case response contents 2 (or 
more) fields WWW-Authentificate like this:

/Server: Apache-Coyote/1.1//
//Vary: Accept-Encoding//
//WWW-Authenticate: Digest realm="1Realm", qop="auth", 
nonce="1485848015167:56976:56060fb913a12670c39d700055b7679d", 
opaque="ca05e8a9b744eb6c91c624ef9c61fb97"//
//WWW-Authenticate: Digest realm="2Realm", qop="auth", 
nonce="1485848015167:96504:777475a2092ab30b6a23349cdb139e75", 
opaque="ca05e8a9b744eb6c91c624ef9c61fb97"//
//Content-Length: 92//
//Content-Type: text/xml;charset=UTF-8/

in this case LWP will use first realm only, so if user provide "2realm" 
credentials this will be ignored by LWP. (bad authorization attempt)

My fix(add to LWP::UserAgent after line 393 in "sub request", before 
"return $class->authenticate"):

/   unless 
($self->{basic_authentication}{$request->uri_canonical->host_port}{$$challenge{realm}}){//
//       $response->header("Client-Warning" =>"Undefined authentication 
realm '$$challenge{realm}'");//
//       next CHALLENGE;//
//   }/

this fix fill compare user defined credentials with server first 
response and use it only if realm name is same.

Ivan Ladygin

```

From olaf@wundersolutions.com on 2017-02-02 20:37:03:
```
On Tue Jan 31 03:43:46 2017, Ladygin@srt-web.com wrote:
> LWP Digest authorisation small bag, in case response contents 2 (or
> more) fields WWW-Authentificate like this:
> 
> /Server: Apache-Coyote/1.1//
> //Vary: Accept-Encoding//
>  //WWW-Authenticate: Digest realm="1Realm", qop="auth",
>  nonce="1485848015167:56976:56060fb913a12670c39d700055b7679d",
> opaque="ca05e8a9b744eb6c91c624ef9c61fb97"//
>  //WWW-Authenticate: Digest realm="2Realm", qop="auth",
[...]

PR #255, opened 2017-04-22
Allow multiple realms on auth

colinnewell:
This is the start of a fix for issue #248.  It's not quite perfect, I've kept the commits separate for now so it's easy to see what's going on.

In principle allowing multiple providers *should* be easy, but there's another feature in the library designed to prevent extra auth attempts that was blocking it from working.  My last commit simply clobbers that feature for now to prove that in principle this can work, but I do want to improve that.

Once this is done it will probably fix issue #201 too.

### DIFF
--- a/lib/LWP/Authen/Basic.pm
+++ b/lib/LWP/Authen/Basic.pm
@@ -48,7 +48,7 @@ sub authenticate
     }
 
     # check that the password has changed
-    my ($olduser, $oldpass) = $ua->credentials($host_port, $realm);
+    my ($olduser, $oldpass);# = $ua->credentials($host_port, $realm);
     return $response if (defined $olduser and defined $oldpass and
                          $user eq $olduser and $pass eq $oldpass);
 
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -417,8 +417,9 @@ sub request {
                         "Unsupported authentication scheme '$scheme'");
                 next CHALLENGE;
             }
-            return $class->authenticate($self, $proxy, $challenge, $response,
+            $response = $class->authenticate($self, $proxy, $challenge, $response,
                 $request, $arg, $size);
+            return $response if $response->is_success;
         }
         return $response;
     }
--- a/t/local/http.t
+++ b/t/local/http.t
@@ -45,7 +45,7 @@ sub _test {
     return plan skip_all => 'We could not talk to our daemon' unless $DAEMON;
     return plan skip_all => 'No base URI' unless $base;
 
-    plan tests => 89;
+    plan tests => 97;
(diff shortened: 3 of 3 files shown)


colinnewell:
Having run some traces of the tests it appears that the extra requests made that the code was guarding against occurred when the authentication failed.  This is the use case I need to focus on.  Of course that's also largely the use case I am working on too.

colinnewell:
I'm starting to think that this isn't actually that check failing, but more that there is state in LWP::UserAgent that isn't getting cleared up after a request is finished and it's causing subsequent test failures.  In my tests I do an invalid request using the credentials "realm1", "user2", "passwd" and in the subsequent test where I have the correct realm2 credentials it is applying those invalid credentials first, then carrying on.  That takes it down a subtly different path in authenticate that then hits the retry prevention.

colinnewell:
That theory appears to be correct.  If I get the authenticate routine to clear up after itself then the code works, but we're back at the situation where we have too many calls.  I am wondering whether the lack of clear up is deliberate so that authentication doesn't need to be done continually.  If that's the case clearing up good credentials is probably bad.  It would probably make sense to clear up invalid ones.  That way we may end up with some odd effects in the case of a dual credential server, but it's probably the least worst option.

colinnewell:
Hmm, this is interesting.  The problem is largely that I'm reusing the request object.  

It appears that the code path I'm hitting is,

- I have auth headers from the previous attempt
- That auth failed, so it then checks to see if the UA password has changed and then will re-apply the new password if so.

I'm trying to figure out if this is a problematic scenario or not.

Certainly if you were to provide credentials for 2 realms, one of which worked I think the code would currently fail.  Well, that's technically what my test is doing, I just didn't realise it.

I can't say I fully understand how the code that checks the new/old password works though which is odd.

colinnewell:
As best I can tell that changed password thing will only work when you have a user agent that implements it's own get_basic_credentials method.  Any other time nothing will change.  In my situation I am using the stock agent so that is the same and we're stuck.

colinnewell:
I think this is about done.  Please say if anything needs changing.

colinnewell:
Looking more closely at issue #201 I don't think this will quite fix it.  I will do a round of real life testing and see what shakes out before declaring it done.  It could do with some review to check I'm not doing anything you're too worried about soon though.

IvanLadygin:
It works to me Colin, thx.

simbabque:
Hi @colinnewell, I know this is old, but would you be willing to rebase it so we can get it merged? Otherwise I can have a look at doing that myself. Thanks.

colinnewell:
Sorry, I've run out of tuits to work on this.  Please do feel free to do whatever you need to with this.

PR #256, opened 2017-04-25
minor changes to make examples work; clarification of external module

dnstandish:
POST example doesn't work with http. rt.cpan.org now requires https.
Explicitly note that HTTP::CookieJar::LWP is a separate install.
Large document example URL no longer valid.  Replace with different site.


oalders:
Thanks @dnstandish!

PR #257, opened 2017-04-27
Added a developer feature to the http.t tests.

colinnewell:
Added a couple of environment variables to allow the tests to be run
manually to help with debugging.  Simply increasing the timeout on the
test server ought to make it easier to debug.  Allowing you to manually
split out the test run so you can run the 2 parts separately ought to
allow you to run them each in the debugger if necessary.

To run them discretely run the server like this,

PERL_LWP_ENV_HTTP_TEST_SERVER_TIMEOUT=10000 perl -I lib t/local/http.t \
daemon

Then the tests like this,

PERL_LWP_ENV_HTTP_TEST_URL=http://127.0.0.1:56957/ perl -I lib \
t/local/http.t

Alternatively to just stop the server from timeing out too soon run the
tests like this,

PERL_LWP_ENV_HTTP_TEST_SERVER_TIMEOUT=10000 perl -I lib t/local/http.t

### DIFF
--- a/t/local/http.t
+++ b/t/local/http.t
@@ -15,7 +15,24 @@ delete $ENV{PERL_LWP_ENV_PROXY};
 $| = 1; # autoflush
 
 my $DAEMON;
+
+# allow developer to manually run the daemon and the tests
+# separately.  Particularly useful for running with the perl
+# debugger.
+#
+# Run the server like this,
+#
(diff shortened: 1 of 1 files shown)


colinnewell:
I've pulled this out of pull request #255 as it's got nothing to do with that pull request, it was just a handy tool to aid with debugging.  It's a very niche thing only really of use to developers working on LWP::UserAgent rather than end users.

oalders:
LGTM

oalders:
Thanks @colinnewell!

PR #259, opened 2017-05-15
revert vars pragma usage removal to not look for external dependencies

thrk2:
Removing vars pragma usage in PR #114 and replacing it with `use base` had unwanted side-effects.

`@ISA` did not look for external dependencies whereas`use base` does.

If the user is in a directory in which he does not have read access and the package does not exist externally, this is an "other error" and Perl aborts:
```stat("./LWP/Protocol/http/SocketMethods.pm", 0x7fff176b5990) = -1 EACCES (Permission denied)```

### DIFF
--- a/lib/LWP/Protocol/http.pm
+++ b/lib/LWP/Protocol/http.pm
@@ -512,6 +512,6 @@ sub increment_response_count {
 #-----------------------------------------------------------
 package LWP::Protocol::http::Socket;
 
-use base qw(LWP::Protocol::http::SocketMethods Net::HTTP);
+use parent -norequire, qw(LWP::Protocol::http::SocketMethods Net::HTTP);
 
 1;

### CODE REVIEW

**Comment by Grinnz** [comment]
Either way this does not require the `vars` pragma.

**Comment by genio** [comment]
Indeed, ```our @ISA = qw(LWP::Protocol::http::SocketMethods Net::HTTP);```
would be preferred to ```use vars```.

**Comment by kbucheli** [comment]
As this seemingly was a remove-legacy-perl-coding-style change, I suggest to add a comment to avoid any future temptations to do it again:
```
# the base or parent pragma is not used here as this leads to unnecessary file lookups which are 
# a) inefficient and 
# b) may cause failures on some restricted environments (e.g. inaccessible working directory)
```

shadowcat-mst:
@thrk2 how do you get into a situation where that's a problem though? Surely if you can't load the components that http.pm requires to work, http.pm won't work, so getting an error now is preferable to silent failure later?

Also, surely if you have http.pm you also have SocketMethods.pm in @INC? This seems like a situation where the gun was already aimed squarely at your own foot, and you probably wanted "no lib '.';" since *lots* of things do require-on-demand of dependencies and making LWP go back to silently failing doesn't seem like a net win.

Note that I'm not a priori saying "this isn't a problem", only that you've got really justified that the case where it *is* is LWP's fault rather than the fault of whoever didn't add "no lib '.'"

shadowcat-mst:
Ah, so, (1) you probably should have "no lib '.'" anyway

(2) SocketMethods.pm should set the %INC entry

reverting to 'our @ISA' is the wrong fix, but yeah, perl crashing trying to load a nonexistent inline package is silly.

kbucheli:
The issue is that there is no SocketMethods.pm. But there is a LWP::Protocol::http::SocketMethods package defined in lib/LWP/Protocol/http.pm [1]. There is nothing to load so there will be no silently ignored failure. This patch only reverts/fixes a regression introduced 3 month ago [2] into LWP.

The point is that the base or parent pragmas first do a 'require' which is not required as there is nothing to load as Net::HTTP loaded at the top and SocketMethods is defined inline. The missing file will always cause an error with 'use parent' and on special circumstances with 'use base' too. As this is Perl, you have several options:
- set @ISA directly
- "use parent -norequire, ..."
- possibly also the %INC suggestion of shadowcat-mst

Whatever solution shall be used, just fix the regression. We run into it as we have a tool which drops privileges by changing user. And if you run it as root with the home directory as usual working directory it fails as . is not accessible any more. Surely the '.' should not be in the library path, but removing that is a bigger project than undoing a regression ;-). Anyway LWP should not depend on such things.

[1] https://github.com/libwww-perl/libwww-perl/blob/master/lib/LWP/Protocol/http.pm#L500
[2] https://github.com/libwww-perl/libwww-perl/commit/aa0c700a93665463ec2e8c0dd2f703faf3eb1396

oalders:
> "use parent -norequire, ..."

This sounds good to me.

oalders:
Thanks @thrk2!

oalders:
@thrk2 would you be able to squash this down to one commit?

oalders:
Thanks @thrk2!

PR #265, opened 2017-09-13
Use tr/// instead of s/// where appropriate

scop:
Not that it would really matter in these particular cases, but `tr` is much faster than `s` in trivial cases like these. And may result in (subjectively) cleaner code.


oalders:
Thanks @scop and @genio!

PR #267, opened 2017-10-20
entirely move away from Getopt::Std; fixes #200

### LINKED ISSUE #200: replace uses of Getopt::Std with Getopt::Long [rt.cpan.org #112317]
Migrated from [rt.cpan.org#112317](https://rt.cpan.org/Ticket/Display.html?id=112317) (status was 'new')

Requestors:
* ether@cpan.org

From ether@cpan.org on 2016-02-23 21:36:08:
```
Getopt::Std was the original very simplistic command-line option processing module. It is now obsoleted by the much more complete solution Getopt::Long, which also supports short options, and is wrapped by modules such as Getopt::Long::Descriptive and Getopt::Long::Modern for simpler usage.

(quote from Perl::Critic::Policy::Freenode::DiscouragedModules)
```

sergeyromanov:
`lwp-mirror` was the last place using `Getopt::Std`.

### DIFF
--- a/bin/lwp-mirror
+++ b/bin/lwp-mirror
@@ -40,7 +40,7 @@ Gisle Aas <gisle@aas.no>
 use strict;
 use warnings;
 use LWP::Simple qw(mirror is_success status_message $ua);
-use Getopt::Std;
+use Getopt::Long qw(GetOptions);
 use Encode;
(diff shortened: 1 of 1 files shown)


oalders:
Thanks @sergeyromanov!

PR #269, opened 2017-11-01
Include unmatched connect error in status string.

eest:
Prior to this change a HTTPS connection would not get additional error
information added to the status string. Compare the $@ contents for the
"Connection refused" case with IO::Socket::INET which was displayed:
```
IO::Socket::INET: connect: Connection refused
```
... and IO::Socket::SSL where the information was lost:
```
Connection refused
```

### DIFF
--- a/lib/LWP/Protocol/http.pm
+++ b/lib/LWP/Protocol/http.pm
@@ -42,6 +42,8 @@ sub _new_socket
 	    $@ =~ /\b(Crypt-SSLeay can't verify hostnames)\b/
 	) {
 	    $status .= " ($1)";
+	} elsif ($@) {
+	    $status .= " ($@)";
 	}
 	die "$status\n\n$@";
     }


oalders:
LGTM

genio:
Me as well

lib/LWP/Protocol/http.pm
```diff
@@ -42,6 +42,8 @@ sub _new_socket
 	    $@ =~ /\b(Crypt-SSLeay can't verify hostnames)\b/
 	) {
 	    $status .= " ($1)";
+	} elsif ($@) {
+	    $status .= " ($@)";
 	}
 	die "$status\n\n$@";
```
skaji: Is `$@` already displayed here?

eest: I am not sure why the $@ string is appended after the newlines at die(). I was displaying the error using the HTTP::Response->status_line() method which properly showed the additional "(Connection refused)" string after this change.

Before:
```
500 Can't connect to example.com:443
```
After:
```
500 Can't connect to example.com:443 (Connection refused)
```

Looking at the parse() function in my HTTP/Response.pm I see this:
```
 25     my($class, $str) = @_;
 26     my $status_line;
 27     if ($str =~ s/^(.*)\n//) {
 28         $status_line = $1;
 29     }
 30     else {
 31         $status_line = $str;
 32         $str = "";
 33     }
```
... so it looks to me like only the part leading up to the first newline is utilized when such a character is present.

skaji: You may see the full error message in content, that is, $res->content.

eest: ... after digging around some more with Carp cluck() I notice that I am not actually using parse(), but instead have as stack looking like this:
```
HTTP::Response::new()
LWP::UserAgent::_new_response()
LWP::UserAgent::send_request()
```

... and in LWP::UserAgent::send_request() the code does this:
```
 201                     my $full = $@;
 202                     (my $status = $@) =~ s/\n.*//s;
 203                     $status =~ s/ at .* line \d+.*//s;  # remove file/line number
 204                     my $code = ($status =~ s/^(\d\d\d)\s+//) ? $1 : &HTTP::Status::RC_INTERNAL_SERVER_ERROR;
 205                     $response = _new_response($request, $code, $status, $full);
```

So similarly to parse() it only keeps the part prior to the newline as the status string.

skaji:
While $@ is always present in content,
yeah, it may be nice to have $@ in status_line too.

oalders:
Thanks for the reviews everyone and thanks for the patch @eest!

Issue #271, opened 2017-11-23
LWP::UserAgent::run_handlers() is undocumented


oalders:
This method gets used internally, but there is no documentation for it at all.  

It should be noted that you cannot correctly run all handlers using this method.  For instance the `response_data` handler requires four arguments, but `run_handlers()` only provides 3.

karenetheridge:
Once it's documented, it needs to be supported that way forever.

Now, changing the internals like that may not be something we intend to do, but we should be mindful of the commitments that we are imposing for ourselves :)

oalders:
I'm happy if we at least start by commenting the code in question.  It looks to me like `response_data` is the only handler with won't run correctly with the current logic.

It looks like at least 4 dists are already using this method: https://grep.metacpan.org/search?q=%5C-%3Erun_handlers&qd=&qft=

PR #274, opened 2017-11-29
fix run_handlers to allow assigning to the request / response

dakkar:
The documentation of the `request_prepare` handler says:

    The method can assign a new request object to $_[0] to replace the
    request that is sent fully.

but that didn't work before this commit.

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -897,16 +897,20 @@ sub handlers {
 
 sub run_handlers {
     my($self, $phase, $o) = @_;
+
+    # were we pass $_[2] to the callbacks, instead of $o, so that they
+    # can assign to it; e.g. request_prepare is documented to allow
+    # that
     if (defined(wantarray)) {
         for my $h ($self->handlers($phase, $o)) {
-            my $ret = $h->{callback}->($o, $self, $h);
+            my $ret = $h->{callback}->($_[2], $self, $h);
             return $ret if $ret;
         }
         return undef;
     }
 
     for my $h ($self->handlers($phase, $o)) {
-        $h->{callback}->($o, $self, $h);
+        $h->{callback}->($_[2], $self, $h);
     }
 }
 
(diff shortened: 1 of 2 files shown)


lib/LWP/UserAgent.pm
```diff
@@ -897,16 +897,20 @@ sub handlers {
 
 sub run_handlers {
     my($self, $phase, $o) = @_;
+
+    # were we pass $_[2] to the callbacks, instead of $o, so that they
```
oalders: Did you mean `here we pass`?

lib/LWP/UserAgent.pm
```diff
@@ -897,16 +897,20 @@ sub handlers {
...
+    # can assign to it; e.g. request_prepare is documented to allow
+    # that
     if (defined(wantarray)) {
         for my $h ($self->handlers($phase, $o)) {
-            my $ret = $h->{callback}->($o, $self, $h);
+            my $ret = $h->{callback}->($_[2], $self, $h);
```
oalders: This is interesting.  I was just looking at this method recently in #271  Does this change the behaviour of the other handlers as well then?

dakkar: It does. The other handlers' documentation says nothing about what happens when you assign to `@_`, and in general I'd expect a Perl programmer to either not touch `@_`, or know that something weird happens when you do.

I see a few ways to resolve the conflict between the documentation and the implementation:

1. change the docs, say that `request_prepare` has to modify the passed-in object
2. apply this change
3. apply a similar change, but have two different `run_handlers` (or an extra parameter) to select if the handlers can assign to `@_`

The main problem with the current situation (and with approach 3): assigning to `@_` in a (for example) `request_prepare` callback will modify the object inside `run_handlers`, the change will be visible to other `request_prepare` callbacks invoked in the same loop, but will disappear once `run_handlers` returns. This is surprising, and hard to debug.

oalders:
Thanks @dakkar and @skaji!

PR #279, opened 2018-02-22
Nawglan/fix cpanfile

nawglan:
cpan-prc challenge 2018.

When trying to use dzil to run tests, it was complaining about missing dependencies.  Looked at the cpanfile and moved some modules from the develop section to the test section, and added a missing dependency to the develop section.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -46,7 +46,10 @@ on 'test' => sub {
     requires 'FindBin';
     requires 'Test::CPAN::Meta';
     requires 'Test::Pod::Spelling::CommonMistakes' => '1.000';
+    requires 'Test::CheckManifest' => '1.29';
+    requires 'Test::Kwalitee'      => '1.22';
     requires 'Authen::NTLM' => '1.02';
+    requires 'Test::CPAN::Changes' => '0.4';
 };
 
 on 'develop' => sub {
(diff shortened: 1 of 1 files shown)


cpanfile
```diff
@@ -53,4 +59,5 @@ on 'develop' => sub {
     requires 'Test::Kwalitee'      => '1.22';
     requires 'Test::Pod::Spelling::CommonMistakes' => '1.000';
     requires 'Try::Tiny'  => '0.24';
+    requires 'Authen::NTLM' => '1.02';
```
oalders: `dzil` generates this file automatically, so it will clobber any changes when it is next run.  You could have a look at https://metacpan.org/pod/distribution/Dist-Zilla/lib/Dist/Zilla/Plugin/Prereqs.pm to see how to handle this via `dist.ini`

nawglan:
Thanks.  On a clean install with a clean version of perl, dzil test, fails because of missing Authen::NTLM.

genio:
@oalders actually, I think we are pulling _from_ the cpanfile, not creating it:

https://github.com/libwww-perl/libwww-perl/blob/master/dist.ini#L40

oalders:
Ah, why *are* we doing that?

karenetheridge:
> On a clean install with a clean version of perl, dzil test, fails because of missing Authen::NTLM.

Can you provide the output you're seeing (which test is failing, and why)?

karenetheridge:
Adding Authen::NTLM to develop-requires prereqs is fine, since it's needed by xt/author/00-compile.t, but the test-requires changes are not needed nor desired.

oalders:
Thanks @nawglan!

PR #281, opened 2018-02-26
use hash access for send_te

### LINKED ISSUE #280: New LWP breaks MSOUTH/ParallelUserAgent-2.62
ParallelUserAgent's test suite started to fail:
```
local/compatibility.t .. 
Failed 1/20 subtests 
local/file.t ........... ok
Oops: Answer (0) for 'http://127.0.0.1:11123/page1' was 
        500: Can't call method "send_te" on unblessed reference at /home/cpansand/.cpan/build/2018022421/libwww-perl-6.32-0/blib/lib/LWP/Protocol/http.pm line 34, <DAEMON> line 1.
 ""
local/http.t ........... 
Failed 8/15 subtests 
local/timeouts.t ....... 
Failed 21/25 subtests 
```
This is very likely due to this change in LWP: https://metacpan.org/source/OALDERS/libwww-perl-6.32/Changes#L5

I think there are two possibilities:
* ParallelUserAgent has to adapt
* Accessing the send_te member in https://metacpan.org/source/OALDERS/libwww-perl-6.32/lib/LWP/Protocol/http.pm#L33 should not be done with a method call, but using hash access. This might sound ugly, but the other two `$self->{ua}` accesses in this function were done just like this, so it would be consistent.

preaction:
This matches with the other hash accessors in this block and should
prevent errors from downstream code that relies on `$self->{ua}` not
being blessed yet.

Fixes #280 

I should mention I have done no digging whatsoever into why this error is happening, but the code was inconsistent: It looks like LWP has always used hash accessors in this constructor, and never assumed that it was a blessed object passed-in.

### DIFF
--- a/lib/LWP/Protocol/http.pm
+++ b/lib/LWP/Protocol/http.pm
@@ -30,7 +30,7 @@ sub _new_socket
 					Proto    => 'tcp',
 					Timeout  => $timeout,
 					KeepAlive => !!$self->{ua}{conn_cache},
-					SendTE    => $self->{ua}->send_te,
+					SendTE    => $self->{ua}{send_te},
 					$self->_extra_sock_opts($host, $port),
 				       );
 


oalders:
Thanks @preaction!

PR #282, opened 2018-03-01
whitespace clean-up: replace tabs with spaces

spazm:
* most of these changes used 8 character tabs
* Ntlm appeared to be using 4 space tabs

#cpan-prc


karenetheridge:
Things like `git blame` become more complicated when having to traverse around stylistic changes. Because this is a very long-lived distribution, and tracking historical changes is important, I lean towards not making this change.

PR #283, opened 2018-03-01
Add response details to timeout() documentation

### LINKED ISSUE #277: LWP::UserAgent Documentation is not clear: timeout is given a 500 error
It took me a while to find it in the code, a `$response` object will have a `$response->code` value of 500 if the connection times out (as well as if the server responds with a genuine _internal server error_

It would be nice to have that more explicit somewhere.

Thankee.

spazm:
fixes #277

#cpan-prc

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,10 +364,15 @@ https-URLs.
 Get/set the timeout value in seconds. The default value is
 180 seconds, i.e. 3 minutes.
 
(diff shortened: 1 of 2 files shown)


karenetheridge:
very useful, thanks!

*thumbs up*

oalders:
Thanks @spazm!

PR #284, opened 2018-03-02
Do not forward the Authentication header after 3xx

### LINKED ISSUE #131: LWP should recalculate digest auth after redirects [rt.cpan.org #24711]
Migrated from [rt.cpan.org#24711](https://rt.cpan.org/Ticket/Display.html?id=24711) (status was 'new')

Requestors:
* oliviert@cpan.org

From oliviert@cpan.org on 2007-02-01 03:04:13:
```
Scenario

1)  client requests / on secure.example.org port 80
2) server replies HTTP/1.1 301 Moved Permanently Location:
http://secure.example/Team
3) client request /Team
4) server replies with 401 challenge (digest auth)
5) client requests /Team again, with auth
6) server replies HTTP/1.1 301 Moved Permanently Location:
http://secure.example/Team/
7) client requests /Team/, with same auth headers

However, the digest authentication response is calculated based (among
other things) on the requested URI, so the response for /Team and /Team/
are different.

Running
% GET -uUsSx http://secure.example.org/Team
returns
[...]
GET http://secure.example.org/Team/
Authorization: Digest username="myusername", realm="REALMNAME",
qop="auth", algorithm="MD5", uri="/Team", nonce=[...]

and so...

8) server say "thanks, but no thanks". 400 Bad Request
logs say:
[Mon Jan 22 01:16:12 2007] [error] [client 133.27.228.213] Digest: uri
mismatch - </Team> does not match request-uri </Team/>


LWP (and, I suppose, LWP::Authen::Digest) should calculate the digest
based on the new URI after a redirect.


---- System info ----

LWP 5.805
perl v5.8.6 built for darwin-thread-multi-2level
Darwin 8.8.0 Darwin Kernel Version 8.8.0: Fri Sep  8 17:18:57 PDT 2006;
root:xnu-792.12.6.obj~1/RELEASE_PPC Power Macintosh powerpc
```

book:
The `Authorization` header is sent when retrying a query after a `401`
response. Usually, a cookie is set, which holds the "authorized" status
of the agent.

If the query including the `Authentication` header is accepted
(i.e. authentication is valid), and the server sends a `3xx` response
back, the original query is cloned, and some headers are removed.

The `Authentication` header must also be removed, otherwise the query
might fail again (with a `403`), for example when the `Location` in
the `3xx` response is a redirect to a different domain, in a different
authorization realm.

Actual debugging and fix by Sergey Belikov.

This should fix #131.

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -321,7 +321,7 @@ sub request {
         my $referral = $request->clone;
 
         # These headers should never be forwarded
-        $referral->remove_header('Host', 'Cookie');
+        $referral->remove_header('Host', 'Cookie', 'Authentication');
 
         if (   $referral->header('Referer')
             && $request->uri->scheme eq 'https'


oalders:
LGTM

oalders:
Should we add a regression test?

vanHoesel:
The change looks clean to me, but I would love to see either RFC that validates this change or have some more test run on modules that depend on us.

> otherwise the query might fail again ...

might or will or only when it rains ?

Can we have some downstream integration test?

PR #285, opened 2018-03-02
add a jigsaw test for 307 redirects

### LINKED ISSUE #42: A redirected POST gets turned into a GET, so no data sent.
This happens even when you add POST to the `requests_redirectable` list.

It has been noted (long ago) in https://rt.cpan.org/Public/Bug/Display.html?id=20727

The problem is caused by this code (`LWP/UserAgent.pl:316` in VERSION 6.04)

```
        if ($code == &HTTP::Status::RC_SEE_OTHER ||
            $code == &HTTP::Status::RC_FOUND)
        {
            my $method = uc($referral->method);
            unless ($method eq "GET" || $method eq "HEAD") {
                $referral->method("GET");
                $referral->content("");
                $referral->remove_content_headers;
            }
        }
```

Changing the unless check to be:

```
            unless ($self->redirect_ok($referral, $response)) {
```

means the test is done against the user-given `requests_redirectable` list

### LINKED ISSUE #129: LWP::UserAgent 302/303 redirection incorrect [rt.cpan.org #20727]
Migrated from [rt.cpan.org#20727](https://rt.cpan.org/Ticket/Display.html?id=20727) (status was 'new')

Requestors:
* cebratcher@gmail.com

From cebratcher@gmail.com on 2006-07-27 19:26:43:
```
When UserAgent->request is called on a page that returns either 302 or
303, the method resets itself to GET.

The following snippet (from the request subroutine) clearly shows the
GET method being set unless the method is GET or HEAD.  


	if ($code == &HTTP::Status::RC_SEE_OTHER ||
	    $code == &HTTP::Status::RC_FOUND) 
        {
	    my $method = uc($referral->method);
	    unless ($method eq "GET" || $method eq "HEAD") {
		$referral->method("GET");
		$referral->content("");
		$referral->remove_content_headers;
	    }
	}


By W3C HTTP/1.1 standards (rfc2616), 302 should not redirect, but rather
prompt the user for redirection.  Now, very very few browsers actually
do this, they just continue on.  However, reformatting the data as GET
is not a viable solution when a host only allows POST.

The 303 should only contain a link and should not redirect at all. 
(Again, most browsers ignore this and just continue on.)

As for a solution...  Simply modifying the routine to check based on the
"requests_redirectable" data rather than what's hardcoded seems like a
logical route to me.


If you need more info or help, feel free to ask!


Thanks!

-Clif
```

From cebratcher@gmail.com on 2006-07-27 20:22:06:
```
Here's a quick fix to replace the previously posted routine.  It
shuffles through the requests_method array and if the method does not
exist in the array, it switches to GET.  A for loop would probably be
faster, but this is a bit more legible.

<code>
	if ($code == &HTTP::Status::RC_SEE_OTHER ||
	    $code == &HTTP::Status::RC_FOUND) 
        {
	    my $method = uc($referral->method);
            my $keep_method = 0;

            foreach (@{$self->requests_redirectable}) {
                if ($method eq $_) { $keep_method++; last; }
            }

	    unless ($keep_method) {
		$referral->method("GET");
		$referral->content("");
		$referral->remove_content_headers;
	    }
	}
</code>


-Clif
```

nalabelle:
This shows that the module follows RFC 7231, sections 6.4.4 and 6.4.7
regarding 307 and 303 redirects. The current behavior is not a bug at
this point.

resolves #42, resolves #129

### DIFF
--- a/xt/author/live/jigsaw/redirect-post.t
+++ b/xt/author/live/jigsaw/redirect-post.t
@@ -1,0 +1,44 @@
+use strict;
+use warnings;
+use Test::More;
+use Test::RequiresInternet ('jigsaw.w3.org' => 80);
+
+use HTTP::Request;
+use LWP::UserAgent;
+use JSON::MaybeXS qw(encode_json);
+use Encode qw(encode_utf8);
+
+plan tests => 10;
+
+my $ua = LWP::UserAgent->new(keep_alive => 1);
+
+my $data = {foo => 'bar', baz => 'quux'};
+my $encoded_data = encode_utf8(encode_json($data));
+
+# 307 not redirectable.
+my $req = HTTP::Request->new('POST', "http://jigsaw.w3.org/HTTP/300/Go_307", undef, undef);
+my $res = $ua->request($req);
+isa_ok($res, 'HTTP::Response', 'request: Got a proper response');
+is($res->code, 307, 'Got a 307 response');
+
+push @{ $ua->requests_redirectable }, 'POST';
+
+# POST can redirect, so support 307 (post to redirected location)
+$res = $ua->request($req);
+isa_ok($res, 'HTTP::Response', 'request: POST redirect got a proper response');
+my $uri = $res->request->uri->as_string;
+my $content = $res->content;
+
+# first we POST to 307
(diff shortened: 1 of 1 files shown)


oalders:
Thanks @nalabelle!

PR #290, opened 2018-04-26
Fix typo in comment ("defaul" -> "default")

anirvan:


oalders:
Thanks for this @anirvan!  I wonder why that trailing space was also in the string?

Issue #288, opened 2018-04-10
Does not support  308 permanent redirect


derobert:
RFC7538 added status code 308 as an additional redirct. 308 is like 301 except without the POST-to-GET possibility.

I think all that'd be needed to add these to LWP is changing (in UserAgent.pm):

```
 if (   $code == HTTP::Status::RC_MOVED_PERMANENTLY
        or $code == HTTP::Status::RC_FOUND
        or $code == HTTP::Status::RC_SEE_OTHER
        or $code == HTTP::Status::RC_TEMPORARY_REDIRECT)
    {
```
to

```
    if (   $code == HTTP::Status::RC_MOVED_PERMANENTLY
        or $code == HTTP::Status::RC_FOUND
        or $code == HTTP::Status::RC_SEE_OTHER
        or $code == HTTP::Status::RC_TEMPORARY_REDIRECT
        or $code == HTTP::Status::RC_PERMANENT_REDIRECT)
    {

```

Not sure which version of HTTP::Status added that constant.

oalders:
Looks like it was added in HTTP-Message 6.07.

https://metacpan.org/diff/file?target=ETHER/HTTP-Message-6.07/lib%2FHTTP%2FStatus.pm&source=GAAS%2FHTTP-Message-6.06%2Flib%2FHTTP%2FStatus.pm

oalders:
So, we just need to bump the version of `HTTP::Status` directly in the `cpanfile`.  PRs are welcome.  :)

oalders:
This should now be fixed via https://metacpan.org/source/OALDERS/libwww-perl-6.38/dist.ini#L43-46

SuneKarlsson:
But the initial issue does not seem to have been taken care of. That LWP does not follow 308 redirects.

galenhuntington:
This is indeed not fixed, because while `HTTP::Status` was bumped, `UserAgent.pm` was not updated.

PR #293, opened 2018-06-05
Don't install HTTP::Status with version 6.17

waterkip:
HTTP::Status 6.16 and 6.17 break the installation of LWP::UserAgent
due to always sending an 'OK' status message even when 404's are being
thrown. This breaks t/local/http.t

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -22,7 +22,7 @@ on 'runtime' => sub {
     requires 'HTTP::Request' => '6';
     requires 'HTTP::Request::Common' => '6';
     requires 'HTTP::Response' => '6';
-    requires 'HTTP::Status' => '6';
+    requires 'HTTP::Status' => '6.18';
     requires 'IO::Select';
     requires 'IO::Socket';
     requires 'LWP::MediaTypes' => '6';
@@ -44,6 +44,7 @@ on 'test' => sub {
     requires 'Test::More';
     requires 'Test::RequiresInternet';
     requires 'FindBin';
+    requires 'HTTP::Status' => '6.18';
 };
 
 on 'develop' => sub {


genio:
6.16 shouldn't present a problem. 6.17 introduced the breaking change.

waterkip:
I'm also happy to bump the required version to 6.18, because https://github.com/libwww-perl/libwww-perl/issues/288 requires a bump to 6.07 as well.

oalders:
The build is failing on `[MakeMaker::Awesome] found version range in runtime prerequisites, which ExtUtils::MakeMaker cannot parse (must specify eumm_version of at least 7.1101): HTTP::Status >= 6, != 6.17`

karenetheridge:
We definitely shouldn't change any prereqs here before HTTP::Status is released again, but perhaps not even then. We can instead make a change in Makefile.PL that upgrades HTTP::Status if it is at the affected version (but only that version - as earlier versions are also fine).

waterkip:
That was my initial fix, but the distzilla plugin Makemaker::Awesome doesn't seem to like the '>= 6, != 6.17' syntax.

karenetheridge:
Just use DynamicPrereqs.

```
[DynamicPrereqs]
:version = 0.018
; Log::Dispatch <= 2.46, >= 2.49
-body = requires('HTTP::Status', '6.18') if has_module('HTTP::Status', '6.17');
```

karenetheridge:
Makemaker::Awesome is fine with the other syntax -- it correctly parses it, and points out that only quite new versions of ExtUtils::MakeMaker can handle it. The other option is to insert extra code into Makefile.PL that runs on the client side and dynamically modifies the prereqs as needed.

karenetheridge:
I have adjusted dist.ini now, and travis is testing it out.

PR #296, opened 2018-07-14
Too Many Open File Handles

genio:
Revert to raw eval in a case where the move to Try::Tiny seems to
have caused a regression where file handles are held open and the
limit for the number of open file handles is reached in a case
where the app requests a file. That file is saved and the req
opened the hostfile and that file remains open.

This behavior happens on every version of Perl lower than 5.18.

### DIFF
--- a/lib/LWP/Protocol.pm
+++ b/lib/LWP/Protocol.pm
@@ -100,73 +100,81 @@ sub collect
     my $content;
     my($ua, $max_size) = @{$self}{qw(ua max_size)};
 
-    try {
+    # This can't be moved to Try::Tiny due to the closures within causing
+    # leaks on any version of Perl prior to 5.18.
+    # https://perl5.git.perl.org/perl.git/commitdiff/a0d2bbd5c
+    my $error = do { #catch
+        local $@;
         local $\; # protect the print below from surprises
-        if (!defined($arg) || !$response->is_success) {
-            $response->{default_add_content} = 1;
-        }
-        elsif (!ref($arg) && length($arg)) {
-            open(my $fh, ">", $arg) or die "Can't write to '$arg': $!";
-	    binmode($fh);
-            push(@{$response->{handlers}{response_data}}, {
-                callback => sub {
-                    print $fh $_[3] or die "Can't write to '$arg': $!";
-                    1;
-                },
-            });
-            push(@{$response->{handlers}{response_done}}, {
-                callback => sub {
-		    close($fh) or die "Can't write to '$arg': $!";
-		    undef($fh);
-		},
-	    });
-        }
-        elsif (ref($arg) eq 'CODE') {
-            push(@{$response->{handlers}{response_data}}, {
-                callback => sub {
-		    &$arg($_[3], $_[0], $self);
-		    1;
-                },
-            });
-        }
-        else {
-            die "Unexpected collect argument '$arg'";
-        }
+        eval { # try
+            if (!defined($arg) || !$response->is_success) {
+                $response->{default_add_content} = 1;
+            }
+            elsif (!ref($arg) && length($arg)) {
+                open(my $fh, ">", $arg) or die "Can't write to '$arg': $!";
+    	    binmode($fh);
+                push(@{$response->{handlers}{response_data}}, {
+                    callback => sub {
+                        print $fh $_[3] or die "Can't write to '$arg': $!";
+                        1;
+                    },
+                });
+                push(@{$response->{handlers}{response_done}}, {
+                    callback => sub {
+                        close($fh) or die "Can't write to '$arg': $!";
+                        undef($fh);
+                    },
+                });
+            }
+            elsif (ref($arg) eq 'CODE') {
+                push(@{$response->{handlers}{response_data}}, {
+                    callback => sub {
+                        &$arg($_[3], $_[0], $self);
+                        1;
+                    },
+                });
+            }
+            else {
+                die "Unexpected collect argument '$arg'";
+            }
 
-        $ua->run_handlers("response_header", $response);
+            $ua->run_handlers("response_header", $response);
 
-        if (delete $response->{default_add_content}) {
-            push(@{$response->{handlers}{response_data}}, {
-		callback => sub {
-		    $_[0]->add_content($_[3]);
-		    1;
-		},
-	    });
-        }
+            if (delete $response->{default_add_content}) {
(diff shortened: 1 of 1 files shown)


simcop2387:
I did a bunch of research on this yesterday and narrowed it down to somehow the $response object is being kept alive.  This ends up leaking file handles for either /etc/hosts or /etc/protocols (it depends on how the connection is made).  Weakening the $response object to break whatever circular ref is made at the start of the sub also fixes this, but is likely to cause other problems.  you can see the results of this research at https://www.simcop2387.info/lwptest/ (it's unorganized, the openstrace-* and trace*.log files are the most useful).  Most testing was done with 5.8.9 or 5.28 once we understood a bit more of what was going on.

genio:
```log
4:35 PM <genio> Recently, a regression in LWP::UserAgent was reported to us. It seems that moving from a raw eval to Try::Tiny caused all Perl versions lower than 5.18 to leave file handles open when grabbing files with the useragent.
4:36 PM <genio> I'm going to do a terrible job at explaining this because I'm running on little sleep, but skaji has also tested the problem as has simcop2387.
4:37 PM <genio> Using Try::Tiny, when a user requests a file multiple times for whatever reason, each file request causes yet another file handle to be left open. Eventually the ulimit -Sn is hit and the UA then sets the x-died header.
4:38 PM <genio> The same exact code, using a raw eval instead, allows the handles to be closed out as expected.
4:38 PM <leont_> That sounds strange
4:40 PM <genio> indeed.  https://github.com/libwww-perl/libwww-perl/pull/296/files?w=1  is the change to LWP::UserAgent that would fix the regression.  However, it begs the question of what might have changed in 5.18 to get around the problem and if there is some way Try::Tiny might be altered to resolve the issue as well
4:43 PM <genio> skaji's gist is fairly small and shows the behavior: https://gist.github.com/skaji/98b53fe081cf57c10ce2b75098f6aa99
4:47 PM <genio> I've got to run away for a bit, but if anyone has any thoughts on what might be causing it, we'd be happy to figure it out
4:53 PM <Bram> best thing to do would be try and get Lwp::UserAgent out of the picture.. that might make it easier to figure out what's going on and/or make bisecting easier
4:54 PM <khw> Thanks to alh, Porting/bisect.pl can bisect modules
5:00 PM <skaji> This script shows Perl 5.16 does not free some resources with Try::Tiny https://gist.github.com/skaji/dfc4a8fdeda03aa1106cb36aa68ae119
5:02 PM <Grinnz> interesting
5:19 PM <Bram> My *guess* is that  https://perl5.git.perl.org/perl.git/commitdiff/a0d2bbd5c  fixed it..  replacing the 'try-catch' block with a closure shows same leak;  unsetting  $res->{callback} also gets rid of the leak..
5:21 PM <Grinnz> oh boy. that wasn't fixed until 5.18
5:21 PM <skaji> ( I slightly simplified the script )
5:22 PM <xenu> yeah the test case from the commit looks *very* similar to skaji's script
5:28 PM <skaji> Bram, Grinnz, xenu: Thanks.
5:29 PM <Bram> Dumping refcount of '$keep' shows that it's 2 (on perl < 5.18).. which means something else is holding a reference to it  (on 5.26.1 refcnt of $keep is 1)
5:31 PM <skaji> Indeed
5:34 PM <leont_> Oh right, I remember that discussion
5:37 PM ⇐ Smylers quit (~smylers@host-78-150-185-177.as13285.net) Quit: Leaving.
5:39 PM <skaji> This problem happens in LWP https://github.com/libwww-perl/libwww-perl/blob/master/lib/LWP/Protocol.pm#L117-L120
5:39 PM <skaji> So what workaround do you recommend?
5:40 PM <Bram> (not only lines 117-120, code before/after also create a closure)
5:50 PM <leont_> Can't think of any obvious solution, it's not easy to move the inner-closure out in this case I think
5:54 PM <Bram> Can the  try  be moved to after setting the callbacks? (i.e. to L135 or so - then again there is still a closure at L140)
5:56 PM <Bram> to me it looks that the only part that can  die  (and thus need try/eval) in upper part of the code is opening of output file.. (so getting rid of that  die  might make moving  try  possible)
6:09 PM <skaji> Bram: Thanks. So your suggestion is like https://github.com/skaji/libwww-perl/commit/61d5d157ef946e3e7de71bb7cbd9418d3dbc9bd4 ?
6:10 PM <Bram> Yes (if it works that is)
6:12 PM <skaji> At least, with that change, I confirmed that LWP did not leave sockets open under perl 5.16.
6:13 PM <Bram> do note that there is still a closure at L148-L150.. so maybe that will leak if the code is reached
6:13 PM ⇐ bjakubski quit (uid70569@id-70569.tooting.irccloud.com) Quit: Connection closed for inactivity
6:26 PM <skaji> If reducing using callbacks eases the problem, should we use plain eval() instead of try()/catch() callbacks by Try::Tiny?
6:26 PM <skaji> And it is exactly genio's PR https://github.com/libwww-perl/libwww-perl/pull/296
6:28 PM <Bram> using a plain eval will not trigger the leak.. but that may have the risk of someone coming along and replacing the eval with a try-catch block later..
6:29 PM <genio> agreed. in which case, we'd need a good test to prevent such a regression
6:30 PM <Bram> That test might be difficult since it'll only leak on < 5.18..
6:35 PM <skaji> We have travis-ci, that tests LWP against perl 5.8 .. 5.26. So we will be able to find a regression if we have a good test case.
6:37 PM <Bram> A test case could be: open a file, record it's fileno, close the file;  call LWP to do a request;  open a file and compare fileno with earlier fileno.. if there are no leaks then it should be the same [unless there is an os that assigns random fileno's when opening files]
6:39 PM <leont_> That's a fragile test, and needlessly complicated
6:39 PM <leont_> The issue isn't about file descriptors, but about leaking variables
6:42 PM <skaji> Oh! we can use Test::LeakTrace for it: https://gist.github.com/skaji/564bf8192fbb35b2e6aff121a685aeba
6:47 PM <xenu> just add a comment
6:47 PM <xenu> also i'm sure that at some point people will stop caring about leaks on perl 5.16
6:48 PM <xenu> rhel 7 will eol in 2024, just 6 more years!
6:51 PM <genio> now with comments: https://github.com/libwww-perl/libwww-perl/pull/296/files?w=1
6:55 PM <xenu> i'm not sure if that's enough, maybe link to that perl commit 
6:58 PM <genio> done
```

oalders:
Thanks everyone!

PR #297, opened 2018-07-15
add no_leak.t

skaji:
Add no_leak.t to ensure LWP does not leak variables.
See #296.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -44,6 +44,8 @@ on 'test' => sub {
     requires 'Test::More';
     requires 'Test::RequiresInternet';
     requires 'FindBin';
+    requires 'Test::Needs';
+    recommends 'Test::LeakTrace';
 };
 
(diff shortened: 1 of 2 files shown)


karenetheridge:
This looks great! Using Test::Needs here with a recommended prereq will ensure we don't block installs when the user can't install the extra dependency, but also let us test on as many platforms as we can.

petdance:
Please add a note in the code about what neverssl.com is, in case that site and the explanation of what it is goes away.  This test is going to be around for a long long time.

skaji:
@genio @karenetheridge @petdance Thank you for your reviews.
I added a note about neverssl.com 0cddd0f

skaji:
Now the PR build (not push build) was passed !
screen shot 2018-07-16 at 12 09 52

karenetheridge:
merged, thanks!

PR #300, opened 2018-10-17
improve handler documentation

### LINKED ISSUE #272: Documentation around `add_handler` needs clarification
The docs look like `response_data => sub { my($response, $ua, $h, $data) = @_; ... }`

My immediate thought when seeing the docs is that `$h` is an `HTTP::Headers` object, but it appears to actually be a handler:

```perl
sub run_handlers {
    my($self, $phase, $o) = @_;
    if (defined(wantarray)) {
        for my $h ($self->handlers($phase, $o)) {
            my $ret = $h->{callback}->($o, $self, $h);
            return $ret if $ret;
        }
        return undef;
    }
 
    for my $h ($self->handlers($phase, $o)) {
        $h->{callback}->($o, $self, $h);
    }
}
```

simbabque:
fixes #272

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 Change history for libwww-perl
 
 {{$NEXT}}
+    - Improve ->add_handler documentation (GH #272) (Julien Fiegehenn)
 
 6.36      2018-10-10 02:20:58Z
     - fix broken link https://metacpan.org/pod/LWP::Simple by fixing pod
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -1596,7 +1596,7 @@ The possible values C<$phase> and the corresponding callback signatures are:
 
 =over
 
-=item response_data => sub { my($response, $ua, $h, $data) = @_; ... }
+=item response_data => sub { my($response, $ua, $handler, $data) = @_; ... }
 
 This handler is called for each chunk of data received for the
 response.  The handler might croak to abort the request.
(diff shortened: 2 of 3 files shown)


genio:
Looks good to me! Thank you for this. 

I’m on my phone so I can’t check at the moment but please ensure the C<< style POD renders correctly. I want to say there needs to be a white space between them and the content within for some reason.

simbabque:
I'm always unsure about those. It might be wrong indeed. Is there a good
way to test this locally?

oalders:
Could you just run it through `perldoc`?

genio:
I usually just do a quick visual inspection using `perldoc lib/LWP/UserAgent.pm` and scroll through to the area to ensure it looks correct. Not a great method, but it works (tm).

simbabque:
I'll do that later or tomorrow, then update. Wait with the merge please.
Sorry about that.

simbabque:
Ok, fixed.

simbabque:
The failed test is a release-only test.

> xt/release/changes_has_content.t ....... Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/2 subtests 

I don't think this test should be run on PRs. It seems to happen with all PRs at the moment.

karenetheridge:
> I don't think this test should be run on PRs. It seems to happen with all PRs at the moment.

Yes, it very much should. Every pull request should have a high level (user-oriented) description of the change, as well as attribution in the form the author prefers.  This makes the release process much smoother.

karenetheridge:
please rebase against the new master (when you come back to this) and `git push --force` -- thanks!

simbabque:
Let me just add Changes... forgot about that.

simbabque:
There we go. All ready to merge. :)

oalders:
Thanks @simbabque!

PR #301, opened 2018-10-18
Issue 186: Change the hostname to use localhost in the tests

### LINKED ISSUE #186: Run tests against localhost instead of hostname [rt.cpan.org #94959]
Migrated from [rt.cpan.org#94959](https://rt.cpan.org/Ticket/Display.html?id=94959) (status was 'new')

Requestors:
* ppisar@redhat.com

Attachments:
* [0001-Connect-to-localhost-instead-of-hostname.patch](https://rt.cpan.org/Ticket/Attachment/1353746/718413/0001-Connect-to-localhost-instead-of-hostname.patch)

From ppisar@redhat.com on 2014-04-23 10:58:43:
```
HTTP::Daemon->new()->url() returns a hostname in the URL. There is no better domain name than the hostname in case of any address.

However that does not mean the hostname is resolvable and reachable over a network. Then the local tests are skipped. In my opinion, localhost is much better identifier in this case.

Attached patch changes local tests to use localhost instead of the hostname.
```

From ppisar@redhat.com on 2015-12-07 08:48:38:
```
Dne St 23.dub.2014 06:58:43, ppisar napsal(a):
> HTTP::Daemon->new()->url() returns a hostname in the URL. There is no
> better domain name than the hostname in case of any address.
> 
> However that does not mean the hostname is resolvable and reachable
> over a network. Then the local tests are skipped. In my opinion,
> localhost is much better identifier in this case.
> 
> Attached patch changes local tests to use localhost instead of the
> hostname.

This was fixed in 6.15 by using "127.0.0.1" host. Still I think "localhost" is better because now it requires IPv4 support and will fail on pure IPv6 hosts.
```

mschae94:
fixes #186

### DIFF
--- a/t/local/http.t
+++ b/t/local/http.t
@@ -466,7 +466,7 @@ sub daemonize {
 
     # Note: tiemout of 0 is not infinite, so no point in special casing
     # timeout logic.
-    my $d = HTTP::Daemon->new(Timeout => $ENV{PERL_LWP_ENV_HTTP_TEST_SERVER_TIMEOUT} || 10, LocalAddr => '127.0.0.1') || die $!;
+    my $d = HTTP::Daemon->new(Timeout => $ENV{PERL_LWP_ENV_HTTP_TEST_SERVER_TIMEOUT} || 10, LocalAddr => 'localhost') || die $!;
     print "Pleased to meet you at: <URL:", $d->url, ">\n";
     open(STDOUT, $^O eq 'VMS'? ">nl: " : ">/dev/null");
 
--- a/t/robot/ua-get.t
+++ b/t/robot/ua-get.t
@@ -121,7 +121,7 @@ sub daemonize {
         exit;  # terminate HTTP server
     };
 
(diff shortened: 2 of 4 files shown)


skaji:
What is the purpose of this PR?
If this is for ipv6 issues, I'm not sure this solves them because HTTP::Daemon uses IO::Socket::INET.

genio:
@skaji I think I agree with your assessment as code stands right now.

However, there is a push to update `HTTP::Daemon`. Downstream (RHEL packagers) have already patched `HTTP::Daemon` to use `IO::Socket::IP` which would use `::INET6` if available. They also have patched a few other things that we need to consider in `HTTP::Daemon` itself.

The `IO::Socket::IP` patch, however, caused a regression reported a while back in IRC that I (unfortunately) can no longer find the test for. I've halted pulling in the downstream patches until I can find that regression test.

oalders:
I've got a rebased version of this PR at #406

PR #302, opened 2018-10-18
travis shouldn't run release tests for pull requests

simbabque:
This is trying to fix the failing travis builds for pull requests. Please ignore until it's working.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,6 @@ before_script:
 script:
   - perl Makefile.PL
   - make
-  - prove -b -r -s -j$(test-jobs) $(test-files)
+  - prove -b -r -s -j$(test-jobs) $(RELEASE_TESTING=$([ "$TRAVIS_PULL_REQUEST" != "false" ] && echo "0" || echo "$RELEASE_TESTING" ) test-files)
 after_success:
   - coverage-report


genio:
This can be corrected by updating the changes file after a release to include a single line with no information:

```
{{$NEXT}}
    - 
```

I think there's value in having the tests run, but we need to ensure they pass after release so that PRs can be tested correctly.

genio:
I updated master. If you rebase your PRs and push, they _should_ pass the xt tests now.

simbabque:
Thanks. I can't do that for a few days. Traveling without computer.
However, shouldn't the next version field be added automatically by another
dzil plugin? Why didn't it?

oalders:
Making the Changes file non-empty but without content could possibly bite me when I run the next release.  That test has caught empty Changes files for me on more than one occasion.  It would be nice to find a solution where this particular test could just get skipped on Travis pull request testing.

karenetheridge:
Release tests shouldn't run for pull requests, but the 'has stuff in Changes' test absolutely *should* run. I'll fix it tonight.

karenetheridge:
Tests are failing for unrelated reasons. It is intentional that pull requests should update the Changes file -- there is code explicitly in dist.ini to add that feature.

simbabque:
> Tests are failing for unrelated reasons. It is intentional that pull requests should update the Changes file -- there is code explicitly in dist.ini to add that feature.

Maybe that could be addressed in the CONTRIBUTING file? It would help future contributors, especially if they are less experienced.

Issue #171, opened 2017-03-31
Client-Warning: Internal response sometimes reset [rt.cpan.org #80601]


oalders:
Migrated from [rt.cpan.org#80601](https://rt.cpan.org/Ticket/Display.html?id=80601) (status was 'new')

Requestors:
* victor@vsespb.ru

From victor@vsespb.ru on 2012-11-03 15:18:06:
```
In the docs there are:

There will still be a response object returned when LWP can't connect to the
server specified in the URL or when other failures in protocol handlers occur.
These internal responses use the standard HTTP status codes, so the responses
can't be differentiated by testing the response status code alone.  Error
responses that LWP generates internally will have the "Client-Warning" header
set to the value "Internal response".  If you need to differentiate these
internal responses from responses that a remote server actually generates, you
need to test this header value.


However this does not work when you set max_redirects to 0.

I am getting the following:


            500 Can't connect to 1.2.3.4:443 (Connection timed
out)

            Content-Type:
text/plain

            Client-Date: Sat, 03 Nov 2012 14:32:56
GMT

            Client-Warning: Redirect loop detected (max_redirect =
0)


(Client-Warning overwrited with another message about redirect loop)
```

PR #303, opened 2018-10-18
Fix Client-Warning: Internal response sometimes reset

augwa:
This fixes issue #171

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -300,7 +300,7 @@ sub request {
     my $response = $self->simple_request($request, $arg, $size);
     $response->previous($previous) if $previous;
 
-    if ($response->redirects > $self->{max_redirect}) {
+    if ($response->header('Location') && $response->redirects >= $self->{max_redirect}) {
         $response->header("Client-Warning" =>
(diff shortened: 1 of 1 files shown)


genio:
I _think_ I remember us discussing this change quite a while back and it presented problems when the option to not follow redirects at all or only follow 1 redirect or something occurs. I can't remember the details off-hand, but let's test a bit.

augwa:
The problem right now is that if the request was successful within the max number of redirects it will complain that it's an redirect loop, even though it's not redirecting at all.

For instance if you asked for 0 max redirects, then EVERY request will always have this even though it never redirected since the redirect count is currently sitting at 0. therefore 0 >= 0 = true.

oalders:
@augwa would you be able to add some tests to this?

simbabque:
I have reproduced this in the most recent master on version 6.43. I was trying to make a test for it so we can get this merged, but I struggle with where to put that test. All the relevant tests are in t/local/http.t, but that uses the local daemon. That seems the wrong place. At the same time, the generic t/base/ua.t looks like it's utterly the wrong place too, and I don't really want to introduce a new file just for this regression test.

oalders:
I think it's fine to introduce a new file for a regression test. Maybe `t/redirect.t`?

PR #308, opened 2019-03-13
Remove s.c.o. from SYNOPSIS

oalders:


genio:
I don't, in any way, disagree with the intent (to no longer point to SCO). But, does anyone object to having an HTTPS URL in the synopsis when that would also require `LWP::Protocol::https` ? I don't know how many non-HTTPS sites we can find these days to point to instead. Does it matter for example purposes?

oalders:
Using `http` might be a game of whack-a-mole, since those URLs are disappearing all the time, but I do get what you're saying.

karenetheridge:
example.com has several endpoints available that are guaranteed to never go
away - because they are intended to serve as examples.

vanHoesel:
yay. thank You @karenetheridge , I was looking for some domain specifically setup for http testing purposes ... but that one is indeed a good 'example' dot com

oalders:
Thanks @karenetheridge.  I've updated the example to use http://example.com

PR #311, opened 2019-03-27
Digest auth expiry module patch

Frank071:
This is essentially the patch that marmotil suggested in pull #40, but updated for the current version of libwww-perl.


Frank071:
Retrying with an all in one patch #313 -- sorry for any inconvenience caused by inexperience...

oalders:
No need to apologize.  We'll get there.  I appreciate your help with this.

PR #314, opened 2019-04-10
Document current best practices

oalders:


lib/LWP/UserAgent.pm
```diff
@@ -1988,6 +2023,47 @@ is in the object's C<requests_redirectable> list,
 false if the proposed redirection is to a C<file://...>
 URL, and true otherwise.
 
+=head1 BEST PRACTICES
+
+The default settings can get you up and running quickly, but there are settings
+you can change in order to make your life easier.
+
+=item cookie_jar
```
genio: Looks like you're using `=item` without a `=over` section here, but you do have a `=back` later on. Would it make more sense to `=head2` these sections so we could have direct URLs for them in metacpan and whatnot later to point people to?

genio:
I'm 👍 on this idea

lib/LWP/UserAgent.pm
```diff
@@ -1133,22 +1133,48 @@ LWP::UserAgent - Web user agent class
...
+    use LWP::UserAgent       ();
 
- if ($response->is_success) {
-     print $response->decoded_content;  # or whatever
- }
- else {
-     die $response->status_line;
- }
+    my $jar   = HTTP::CookieJar::LWP->new;
+    my $agent = LWP::UserAgent->new(
```
skaji: s/$agent/$ua/

Also, I think we should use the same variable name for LWP::UserAgent objects in this document.
So why don't you use `$ua` in "BEST PRACTICES" section too?

lib/LWP/UserAgent.pm
oalders: There already is a `=head2 cookie_jar`, so that might not work here. I'm open to some other way of doing this, though.

lib/LWP/UserAgent.pm
```diff
@@ -1133,22 +1133,48 @@ LWP::UserAgent - Web user agent class
...
+        protocols_allowed => ['http', 'https'],
+        timeout           => 10,
+    );
+
+    $ua->env_proxy;
+
+    my $response = $ua->get('http://example.com');
+
+    if ($response->is_success) {
+        print $response->decoded_content;    # or whatever
```
vanHoesel: It can do without the `# or whatever` comment

I usually use the `...` to indicate arbitrary code that can be anything before or after ... or here, in the middle

 But having the `decoded_content` is a good example too

lib/LWP/UserAgent.pm
vanHoesel: 🤔 and no pod checker that was complaining ?

oalders: There was a failing test that Travis caught.  I just didn't run my tests locally.  :(

lib/LWP/UserAgent.pm
oalders: Thanks, I've fixed all of those cases now.  This was partly a copy/paste from `WWW::Mechanize`, which is where `$agent` came from.  Although, looking more closely at that, it should probably have been `$mech` in the `WWW::Mechanize` docs, since that's what everything else there uses.

lib/LWP/UserAgent.pm
genio: Might I propose a different name to not conflict with `cookie_jar`? Since this is a Best Practices thing, it can be named anything we like. Maybe `=head2 Handling Cookies` or some such?

lib/LWP/UserAgent.pm
genio: I agree, I'm not sure why someone *cough* *cough* ever put that in there. It reads oddly passive aggressive. /me hides

PR #315, opened 2019-04-11
Document order in which handlers fire

oalders:


genio:
Looks good to me!

oalders:
Thanks @genio!

PR #318, opened 2019-05-13
Add the patch method to LWP::UserAgent.

### LINKED ISSUE #116: New Feature: HTTP PATCH
Hi
Please add support the HTTP PATCH verb.
Probably very similar to put :)
Thanks
Andrew

genio:
Much like the previously added `put` function, the `patch` function will
require a specific version of HTTP::Request::Common. This time, though,
we will need at least version 6.12.

This is largely a copy-paste of the already added `put` function. All of
the current tests we have in place that test the `put` funtion were
duplicated for the new `patch` function. The same logic gate that only
tests when we have the proper version of HTTP::Request::Common was added
for the 6.12 check.

This makes LWP::UserAgent a bit simpler to work with RESTful endpoints
in that we now directly support the most common HTTP verbs: `HEAD`,
`GET`, `DELETE`, `PUT`, `POST`, `PATCH`.

Fixes #116

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 Change history for libwww-perl
 
 {{$NEXT}}
+    - Add the patch method (GH#116)
 
 6.39      2019-05-06 14:18:39Z
     - Document current best practices (GH#314) (Olaf Alders)
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -490,6 +490,22 @@ sub head {
 }
 
 
+sub patch {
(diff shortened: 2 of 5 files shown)


genio:
There's a stupid fix in there for now. I need to clean this up more before thoughts of merging.

oalders:
This now has merge conflicts.

genio:
I need to redo this, I think. The web based conflict manager thingy breaks the timeline like it's its job

oalders:
Continued in #334

PR #319, opened 2019-05-22
Fix documentation for removing all handlers

leedo:
This example references a non-existent method `remove_handlers`. It should call `remove_handler` with no arguments to remove all handlers instead.


oalders:
Thanks @leedo!

PR #320, opened 2019-06-20
Fix threaded Perl tests.

genio:
Some threaded versions of Perl hit an odd failure on the
xt/author/live/jigsaw/auth-d.t tests.

Failing back to an empty array ref when there's nothing left to grab
from the test set masks the problem. What we haven't yet had time to
investigate is _why_ the tests are asking for more data than what we've
provided. There's some odd quirk going on that we should investigate and
figure out when time permits. THIS IS A BANDAID.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,9 +44,6 @@ matrix:
       perl: '5.28'
 env:
   global:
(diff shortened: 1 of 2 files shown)


genio:
This is useless.

PR #313, opened 2019-03-27
Let Digest authentication act on nonce expiry

Frank071:
Back in 2013 marmotil discovered that libwww-perl did not correctly act on the expiry of nonce as part of Digest authentication. For some reason this was not picked up on. This pull request essentially contains the original patches made by marmotil, but adapted to the current version.

Pull requests #40, #94, #311 and #312 are obsolete after applying this.

PR #321, opened 2019-06-20
Squashed and rebased pr 313

oalders:
This is a re-worked version of #313.  Mostly just messing with Git history and then tidying.


oalders:
I'm going to merge this.  It has already had one code review and the only real changes since then have been rebases.

PR #324, opened 2019-09-28
doc: fields starting with ":" have to be quoted

eserte:


oalders:
Thanks @eserte and @skaji!

PR #326, opened 2019-10-09
Make file arg for mirror mandatory #304

### LINKED ISSUE #304: LWP::UserAgent::mirror() should do some basic parameter validation
```perl
use strict;
use warnings;

use Devel::Confess;
use LWP::UserAgent;

my $ua = LWP::UserAgent->new;
$ua->mirror('http://cpancover.com/latest/cpancover.json');
```

```sh
$ time perl mirror.pl
Cannot rename '-76047' to '': No such file or directory
 at /Users/olafalders/.plenv/versions/5.26.1/lib/perl5/site_perl/5.26.1/LWP/UserAgent.pm line 1009.
	LWP::UserAgent::mirror(LWP::UserAgent=HASH(0x7fe5f8002690), "http://cpancover.com/latest/cpancover.json") called at mirror.pl line 8

real	1m3.107s
user	0m0.926s
sys	0m0.428s
```

`LWP::UserAgent::mirror()` never checks to see if the second argument to `mirror()` (the file name) is even defined.  In this case I had to wait for a 17MB file to download in order to find out that something was wrong.  The default error message (without `Devel::Confess`) is even less helpful.

simbabque:
The ticket talks about "basic parameter validation". I think we only need to make sure the second arg is there. No need to check the URL as that will be done when we GET the file.

Closes #304

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -966,6 +966,8 @@ sub mirror
 {
     my($self, $url, $file) = @_;
 
+    die "Local file name is missing" unless $file;
+
     my $request = HTTP::Request->new('GET', $url);
 
     # If the file exists, add a cache-related header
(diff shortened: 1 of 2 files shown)


oalders:
Thanks @simbabque!

2shortplanks:
@oalders We're not allowing mirroring to files called `0` now?

oalders:
@2shortplanks good point.  @simbabque maybe this should be a check for defined?

2shortplanks:
Opened issue #329 and created pull request

simbabque:
> @2shortplanks good point. @simbabque maybe this should be a check for defined?

Yes, it should. My bad. Sorry for the late response, had no access to a computer and then lost track of this.

PR #329, opened 2019-10-28
Allow mirroring to files called '0'

2shortplanks:
After GH#326 we were no longer able to mirror to files called '0'

Now we check for length and definedness rather than truthfullness

And I added tests


2shortplanks:
Addresses issue #330

oalders:
Thanks @2shortplanks! This now lives at https://metacpan.org/release/OALDERS/libwww-perl-6.41

PR #332, opened 2019-11-13
Set up unlimited conn cache on keep_alive => undef

scop:
Matches implementation with the docs, which say the cache is created
based on whether keep_alive is passed in, not whether it's defined.

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -72,10 +72,11 @@ sub new
 
     my $cookie_jar = delete $cnf{cookie_jar};
     my $conn_cache = delete $cnf{conn_cache};
+    my $keep_alive_exists = exists $cnf{keep_alive};
     my $keep_alive = delete $cnf{keep_alive};
 
     Carp::croak("Can't mix conn_cache and keep_alive")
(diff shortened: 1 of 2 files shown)


skaji:
In this case, I think we should fix docs instead of code.

simbabque:
I agree with @skaji on this one. I've tracked down the documentation in question back to [this commit](https://github.com/libwww-perl/libwww-perl/blame/a224a0079c5379b9ca394784474a3fcb3ac8c242/lib/LWP/UserAgent.pm#L172), which is 19 years ago. The functionality was introduced at that point too, and has basically been around unchanged ever since. I strongly feel we should not temper with that, as it might break a lot of things, especially since we would `die` in additional cases.

oalders:
@scop thanks very much for your work on this.  Would you be willing to change this PR over to a documentation fix?

oalders:
Maybe the test could be changed to conform to the current behaviour as well.

scop:
Sure, #333 has the fix the other way around.

oalders:
Closing as work has moved to #333.  Thanks @scop, @skaji  and @simbabque!

PR #333, opened 2019-11-15
Fix docs to match keep_alive => undef behavior, add some trivial tests

scop:
As discussed in PR #332


skaji:
@scop Thank for this.

`keep_alive => 0` does not setup conn_cache, either. 
```
my $ua = LWP::UserAgent->new(keep_alive => 0);
is $ua->conn_cache, undef;
```

So how about changing
```
If keep_alive option value is defined,
```
to
```
If keep_alive option value is defined and > 0,
```
?

scop:
I chose to go with "defined and non-zero" and added a related test. If you wish to tweak this further, my branch is open for maintainer commits, be my guest :)

oalders:
Thanks @scop and @skaji!

PR #335, opened 2019-11-21
FIX: continue challenge schemes

simbabque:
Original code by @romangaleev. I've added tests. See  #78 for details.

Closes  #78.


oalders:
Thanks @romangaleev, @simbabque and @genio!

PR #338, opened 2020-02-03
Add blank line after POD heading for patch method

Eckankar:
Omitting this blank line makes certain POD parsers believe that the
following line is still a part of the heading. This can be seen on
metacpan.org, for instance:

image


genio:
@Eckankar Thanks! Good catch.

PR #339, opened 2020-04-12
Support basic authentication charset per RFC 7617

scop:
https://tools.ietf.org/html/rfc7617#section-2.1

### DIFF
--- a/t/local/http.t
+++ b/t/local/http.t
@@ -63,7 +63,7 @@ sub _test {
     return plan skip_all => 'We could not talk to our daemon' unless $DAEMON;
     return plan skip_all => 'No base URI' unless $base;
 
-    plan tests => 109;
+    plan tests => 125;
 
     my $ua = LWP::UserAgent->new;
     $ua->agent("Mozilla/0.01 " . $ua->agent);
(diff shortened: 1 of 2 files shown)


t/local/http.t
```diff
@@ -246,6 +246,31 @@ sub _test {
         isa_ok($res, 'HTTP::Response', 'basicAuth: good response object');
         is($res->code, 401, 'basicAuth: code 401');
     }
+    { # basic auth, UTF-8
+        my $req = HTTP::Request->new(GET => url("/basic_utf8", $base));
+        my $res = MyUA4->new->request($req);
+        isa_ok($res, 'HTTP::Response', 'basicAuthUtf8: good response object');
```
vanHoesel, line 252: is it possible to rename this (and others) to `'basicAuthUTF8'`, because `'basicAuthUtf8'` is hard to read

t/local/http.t
```diff
@@ -451,6 +487,22 @@ sub daemonize {
...
+        if (defined($u) && utf8::decode($u) && utf8::decode($p) && $u eq 'ök 12' && $p eq 'xyzzy ÅK€j!') {
+            $c->send_basic_header(200);
+            $c->print("Content-Type: text/plain");
+            $c->send_crlf;
+            $c->send_crlf;
+            $c->print("$u\n");
+        }
+        else {
+            $c->send_basic_header(401);
+            $c->print("WWW-Authenticate: Basic realm=\"libwww-perl-utf8\", charset=\"UTF-8\"\015\012");
```
vanHoesel, line 502: this returns the uppercase variant , what if I were to return a lowercase ? that is not covered by test

t/local/http.t
scop: Matter of taste, but I see your suggestion is more in line with what LWP otherwise uses, so changed.

t/local/http.t
scop: We could add that, but I think it's kind of trivial, and I believe cases like returning `www-authenticate` and `basic` in lowercase or `REALM` in non-lowercase are not tested either.

vanHoesel: That other things in other parts of te code are not tested either, is not a valid argument to do it yourself, is it ?

I was not referring to the `'charset'` name itself, but more to the value assigned to it. I rather not overlook the 'dark magic' happening under the hood:

The RFC mention this as to be 'case-insensitively' to be "UTF-8", this is (reading the code) checked, using `uc`, and then relying on the `Encode::encode` routine to both understand lower-case and upper-case (or even mixed-case).
What if `encode` would not do its magic?

Tests are to ensure that things are done right, even though I can read code.

oalders:
Thanks for this @scop! Looks like the coverage goes up with this branch, which is nice.  I think allowing for the trivial test cases is nice as well. We need as much coverage as we can get. 😄

t/local/http.t
scop: Existing code/tests serve as precedent, just as they did for the naming thing.

Do you think a test that repeats the same as the current one, just with `utf-8` would suffice, or would you like mixtures like `Utf-8`, 'uTf-8` etc tested as well?

scop: In the meantime, modified so that we only ever pass uppercase 'UTF-8` to `encode'.

vanHoesel: I am totally fine with testing agains 'UTF-8' and 'utf-8'.

I will not ask you to do negative testing against 'UTF8' or 'utf8', the latter being something totally different for `encode`

scop: Ok, done. I still kind of fail to see the point though -- to me after the "uppercase only to encode" change this is pretty much equivalent to testing whether `uc($something) eq "UTF-8"`, and I don't know why we'd want to test something like that in LWP. But guess it serves as a guard for some possible future changes. But nevermind.

Sanitizing the `encode` arg was a good thing to do, thanks for pointing that out.

oalders:
Thanks @scop and @vanHoesel!

oalders:
https://metacpan.org/release/OALDERS/libwww-perl-6.44

PR #341, opened 2020-06-03
Fix Client-Warning: Internal response sometimes reset (fixup)

### LINKED ISSUE #171: Client-Warning: Internal response sometimes reset [rt.cpan.org #80601]

simbabque:
Rebased work by @augwa onto current master and added tests.

Fixes #171 
Closes #303

### DIFF
--- a/t/redirect.t
+++ b/t/redirect.t
@@ -1,0 +1,41 @@
+use strict;
+use warnings;
+use Test::More;
+
+use LWP::UserAgent;
+plan tests => 4;
+
+# This is a regression test for #171
+
+my $ua = LWP::UserAgent->new();
+
+{ # default number of redirects
+    $ua->timeout(1);
+    my $res = $ua->get('http://10.255.255.1/');
+    like(
+        $res->header("Client-Warning"),
+        qr/Internal Response/i,
+        'Timeout gives a client warning'
+    );
+    like(
+        $res->content,
+        qr/Can't connect/i,
+        '... and has tells us about the problem'
+    );
+}
+
+{ # no redirects
+    $ua->timeout(1);
+    $ua->max_redirect(0);
+    my $res = $ua->get('http://10.255.255.1/');
+    like(
+        $res->header("Client-Warning"),
+        qr/Internal Response/i,
+        'Timeout with no redirects gives a client warning'
+    );
+    like(
+        $res->content,
+        qr/Can't connect/i,
+        '... and has tells us about the problem'
+    );
+}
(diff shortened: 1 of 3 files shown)


simbabque:
Sorry about that. I keep forgetting the Changes.

t/redirect.t
```diff
@@ -0,0 +1,41 @@
...
+use LWP::UserAgent;
+plan tests => 4;
+
+# This is a regression test for #171
+
+my $ua = LWP::UserAgent->new();
+
+{ # default number of redirects
+    $ua->timeout(1);
+    my $res = $ua->get('http://10.255.255.1/');
```
genio, line 14: `10.255.255.1` is a private address. Some routers use this as their default setup address. This could actually respond on someone's home network.

Is there a better address to use for this purpose?

karenetheridge:
There are specific blocks of IP addresses reserved for testing purposes. Please use one from there.  https://stackoverflow.com/questions/10456044/what-is-a-good-invalid-ip-address-to-use-for-unit-tests

simbabque:
> There are specific blocks of IP addresses reserved for testing purposes. Please use one from there. https://stackoverflow.com/questions/10456044/what-is-a-good-invalid-ip-address-to-use-for-unit-tests

I was looking for that, but came across a similar question on SO with an answer that was much higher voted and had the IP I ended up using. This makes a lot more sense. I'll change it. Thanks.

simbabque:
> @simbabque The other one should be changed, too, shouldn't it?

Sorry, it's not my day.

oalders:
Thanks @augwa and @simbabque!

PR #342, opened 2020-06-15
Fix a looping regression for 303 responses without a Location header

ntyni:
The change in 6.45 had a side effect of making it look at max_redirect only when the Location header is set.

The libapache2-mod-perl2 test suite implements a test server that answers with HTTP code 303 without including such a header.

RFC 7231 is not very clear on whether Location is required, although
the response doesn't make much sense without it.

Some googling reveals this is not a new question.

  https://stackoverflow.com/questions/16194988/for-which-3xx-http-codes-is-the-location-header-mandatory

For the sake of robustness I suppose libwww-perl needs to be fixed.

Bug-Debian: https://bugs.debian.org/962904

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,8 @@
 Change history for libwww-perl
 
 {{$NEXT}}
+    - Fix a looping regression in 6.45 for redirect responses without
+      a Location header (GH PR#342, Niko Tyni)
 
 6.45      2020-06-08 14:51:28Z
     - Fix Client-Warning: Internal response sometimes reset (GH#341) (Jonathan
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -300,10 +300,12 @@ sub request {
     my $response = $self->simple_request($request, $arg, $size);
     $response->previous($previous) if $previous;
 
-    if ($response->header('Location') && $response->redirects >= $self->{max_redirect}) {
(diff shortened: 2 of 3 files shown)


ntyni:
Not sure if this reintroduces the issue 6.45 was fixing. The t/redirect tests still pass, though.

oalders:
@ntyni are you able to have a look at `xt/author/live/jigsaw/redirect-post.t`? It's failing on Travis right now.

Changes
```diff
@@ -1,6 +1,8 @@
 Change history for libwww-perl
 
-{{$NEXT}}
```
oalders, line 3: This is actually a template variable that `Dist::Zilla` will be looking for.  Lines 4-5 are what we need here.

ntyni: Thanks, that was silly of me. Fixed.

ntyni:
> @ntyni are you able to have a look at `xt/author/live/jigsaw/redirect-post.t`? It's failing on Travis right now.

I doubt it regressed with this. It's working for me now. Not sure if it's intentional that only some of the live/jigsaw/ tests are skipped for NO_JIGSAW=1 ?

oalders:
One of the issues with these live tests is that they can start to fail if there's some throttling going on. I can see that it's now passing, so we can punt on it for now! :)

PR #347, opened 2020-08-17
Only add cookies from the jar if there aren't cookies in the request

### LINKED ISSUE #5: Re-using HTTP::Request objects can cause cookie duplication
Was running into issues with an existing script on a machine using LWP::UserAgent 5.834 that was previously working fine under a very old LWP::UserAgent 2.33 environment.

This script starts out by creating several HTTP::Request objects, a cookie jar using HTTP::Cookies, and an HTTP::UserAgent. The UserAgent object has the cookie_jar parameter set with my HTTP::Cookies object.

It then enters a loop that re-uses the HTTP::Request objects every time the loop runs. (Calling $ua->request($req) on each request object to be re-used.)

I was doing this to try and be more efficient, not wanting to call new HTTP::Request multiple times for every run through the loop.

Well, under this new version of LWP::UA, I found that my outgoing request headers were doubling after every request. It turns out that every time UA calls the request method, it ends up running handlers. And since the UA object has a cookie jar set, one of the handlers that runs calls the add_cookie_header() method on the given request object. 

I was able to workaround this issue by not setting the cookie_jar parameter at all, and instead, calling add_cookie_header() on each request object, one time, at the top of my script. 

I don't know if this is a known issue, but it took me several hours to track down this problem so I figured I would bring it up.

Perhaps the Cookies object could be re-written so that it doesn't call add_cookie_header if the request object already has that header set? Or something equivalent?

oalders:
This commit is basically a copy of @preaction's work. I just had trouble
cherry-picking his commit into the repository.

Closes #5

### DIFF
--- a/.github/workflows/dzil-build-and-test.yml
+++ b/.github/workflows/dzil-build-and-test.yml
@@ -28,7 +28,7 @@ jobs:
           NO_JIGSAW: 1
           SENDMAIL: "/usr/sbin/sendmail" # for compilation test of LWP::Protocol::mailto
         run: upgrade-perl-helpers && auto-build-and-test-dist
-      - uses: actions/upload-artifact@master
+      - uses: actions/upload-artifact@v2
         with:
           name: build_dir
           path: build_dir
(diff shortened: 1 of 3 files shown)


oalders:
Thanks @genio!

PR #349, opened 2020-09-15
Support 308 Permanent Redirect.

### LINKED ISSUE #288: Does not support  308 permanent redirect

galenhuntington:
Closes #288 (for real).  Although the dependency version was bumped, the fix wasn't complete because this file wasn't updated.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,8 +1,6 @@
 Change history for libwww-perl
 
 {{$NEXT}}
-
-6.48      2020-09-15 21:58:57Z
     - Support 308 Permanent Redirect (GH#349) (Galen Huntington)
 
 6.47      2020-08-18 15:27:27Z
--- a/Changes
+++ b/Changes
@@ -2,6 +2,9 @@ Change history for libwww-perl
 
 {{$NEXT}}
 
+6.48      2020-09-15 21:58:57Z
+    - Support 308 Permanent Redirect (GH#349) (Galen Huntington)
+
 6.47      2020-08-18 15:27:27Z
     - Only add cookies from the jar if there aren't cookies in the request
       (GH#347) (Doug Bell and Olaf Alders)
--- a/.github/workflows/dzil-build-and-test.yml
+++ b/.github/workflows/dzil-build-and-test.yml
@@ -77,6 +77,12 @@ jobs:
           - os: windows-latest
             perl-version: "5.12"
           - os: windows-latest
+            perl-version: "5.14"
+          - os: windows-latest
+            perl-version: "5.16"
+          - os: windows-latest
+            perl-version: "5.18"
+          - os: windows-latest
             perl-version: "5.32"
     name: Perl ${{ matrix.perl-version }} on ${{ matrix.os }}
     steps:
--- a/.github/workflows/dzil-build-and-test.yml
+++ b/.github/workflows/dzil-build-and-test.yml
@@ -77,12 +77,6 @@ jobs:
           - os: windows-latest
             perl-version: "5.12"
           - os: windows-latest
-            perl-version: "5.14"
-          - os: windows-latest
-            perl-version: "5.16"
-          - os: windows-latest
-            perl-version: "5.18"
-          - os: windows-latest
             perl-version: "5.32"
     name: Perl ${{ matrix.perl-version }} on ${{ matrix.os }}
     steps:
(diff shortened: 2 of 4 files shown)


oalders:
Thanks for his @galenhuntington! The PR just needs an appropriate entry in `Changes` in order for the build to run to completion.

galenhuntington:
Hm, still failing, does `META.json` need to be updated or something like that?

Changes
```diff
@@ -2,6 +2,9 @@ Change history for libwww-perl
 
 {{$NEXT}}
 
+6.48      2020-09-15 21:58:57Z
```
karenetheridge, line 5: This line (and the blank line before it) should be removed; the version and timestamp and filled in from the `{{$NEXT}}` template.

karenetheridge (changes requested):

oalders:
This is failing on some older Perls:

```
t\00-report-prereqs.t .. ok
##[error]Bareword "HTTP::Status::RC_PERMANENT_REDIRECT" not allowed while "strict subs" in use at D:\a\libwww-perl\libwww-perl\lib/LWP/UserAgent.pm line 318.
##[error]Compilation failed in require at t\10-attrs.t line 5.
##[error]BEGIN failed--compilation aborted at t\10-attrs.t line 5.
t\10-attrs.t ........... 
Dubious, test returned 255 (wstat 65280, 0xff00)
No subtests run 
##[error]Bareword "HTTP::Status::RC_PERMANENT_REDIRECT" not allowed while "strict subs" in use at D:\a\libwww-perl\libwww-perl\lib/LWP/UserAgent.pm line 318.
##[error]Compilation failed in require at t\redirect.t line 5.
##[error]BEGIN failed--compilation aborted at t\redirect.t line 5.
t\redirect.t ........... 
```

galenhuntington:
Do old versions of Perl somehow pull in old versions of `HTTP::Status`?  I must not understand the build process.

galenhuntington:
It's only old Perls on Windows too.

galenhuntington:
I don't know if it's an acceptable solution, but dropping the three pre-5.20 Windows Perl versions got us through.

oalders:
So, in https://github.com/libwww-perl/libwww-perl/runs/1120556913?check_suite_focus=true `HTTP::Status` is not getting upgraded. I guess this is because it meets the requirements in https://github.com/libwww-perl/libwww-perl/blob/master/cpanfile#L25

So, in that failing version we have `HTTP::Status             6.03`, which satisfies `requires 'HTTP::Status' => '6';`

We have in `dist.ini`:

```
[DynamicPrereqs]
:version = 0.018
; HTTP::Status 6.17 was buggy
-body = requires('HTTP::Status', '6.18') if has_module('HTTP::Status', '6.17');
```
but that only deals with the case where `6.17` has been installed. It looks like we need to bump the minimum version in `cpanfile`.

oalders:
I took a closer look at all of this. We've been trying to be pretty conservative with bumping prereqs. I think bumping to 6.07 rather than 6.18 would fix this as well. Could we try that? https://metacpan.org/changes/distribution/HTTP-Message#L92-103

oalders:
🚀 

Thanks @galenhuntington! I'm going to squash these and commit them from the command line, so I will close this shortly, once I've released a new version. Thanks so much for this!

oalders:
https://metacpan.org/pod/release/OALDERS/libwww-perl-6.48/lib/LWP/UserAgent.pm

PR #351, opened 2020-09-23
Require network testing enabled for t/redirect.t

### LINKED ISSUE #350: Redirect tests fail in environment with web gateway
The tests in `t/redirect.t` introduced in issue #171 fail when run from our company network. This seems to be because the web gateway immediatly acceps connections on the used test IP `198.51.100.123`. So the `$res->content` result actually contains a read timeout. While this is an issue with the web gateway, I think this could affect other users aswell and it might not be easily fixable in the gateway. Instead it might be simpler to modify the test to handle this special environment. Imho the exact `$res->content` result does not matter for the test.

The exact test output is:
```
#   Failed test '... and has tells us about the problem'
#   at t/redirect.t line 20.
#                   'read timeout at /usr/local/share/perl/5.22.1/Net/HTTP/Methods.pm line 268.
# '
#     doesn't match '(?^i:Can't connect)'
#   Failed test '... and has tells us about the problem'
#   at t/redirect.t line 36.
#                   'read timeout at /usr/local/share/perl/5.22.1/Net/HTTP/Methods.pm line 268.
# '
#     doesn't match '(?^i:Can't connect)'
```

oalders:
Fixes #350

### DIFF
--- a/t/redirect.t
+++ b/t/redirect.t
@@ -1,17 +1,36 @@
 use strict;
 use warnings;
-use Test::More;
-
-use LWP::UserAgent;
-plan tests => 4;
 
 # This is a regression test for #171
 
+use Test::More;
+
+# Test::RequiresInternet is used here basically just to SKIP tests if
+# NO_NETWORK_TESTING has been enabled. We would want to do this particularly if
+# there is a badly behaved router on the network where the tests are being run.
+use Test::RequiresInternet;
+
+use LWP::UserAgent ();
+
+# Regarding the choice of 198.51.100.123 as a test IP address, please see
+# https://tools.ietf.org/html/rfc5737
+#
+# The RFC contains the following description for the block to which this
+# address belongs:
+#
+# Documentation Address Blocks
+#
+# The blocks 192.0.2.0/24 (TEST-NET-1), 198.51.100.0/24 (TEST-NET-2), and
+# 203.0.113.0/24 (TEST-NET-3) are provided for use in documentation.
+
+my $url = 'http://198.51.100.123/';
+
 my $ua = LWP::UserAgent->new();
 
-{ # default number of redirects
+# default number of redirects
+{
     $ua->timeout(1);
-    my $res = $ua->get('http://198.51.100.123/');
+    my $res = $ua->get($url);
     like(
         $res->header("Client-Warning"),
         qr/Internal Response/i,
@@ -24,10 +43,11 @@ my $ua = LWP::UserAgent->new();
     );
 }
 
-{ # no redirects
+# no redirects
+{
     $ua->timeout(1);
     $ua->max_redirect(0);
-    my $res = $ua->get('http://198.51.100.123/');
+    my $res = $ua->get($url);
     like(
         $res->header("Client-Warning"),
         qr/Internal Response/i,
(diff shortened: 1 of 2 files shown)


t/redirect.t
```diff
@@ -1,15 +1,15 @@
 use strict;
 use warnings;
-use Test::More;
 
-use LWP::UserAgent;
-plan tests => 4;
+use Test::More;
+use Test::RequiresInternet;
```
karenetheridge, line 5: Should this be `use Test::RequiresInternet '198.51.100.123';` ?

(Also, what is this IP? I know, from previous discussions in PRs, but it is not explained in the tests. If it has specific behaviour that the test expects, that should be explicitly stated in case we ever have to find an equivalent replacement.)

oalders:
```
0$ git diff
diff --git a/t/redirect.t b/t/redirect.t
index 8aa8c09b..004f29f5 100644
--- a/t/redirect.t
+++ b/t/redirect.t
@@ -2,7 +2,7 @@ use strict;
 use warnings;

 use Test::More;
-use Test::RequiresInternet;
+use Test::RequiresInternet '198.51.100.123';
 use LWP::UserAgent ();

 # This is a regression test for #171
```

```
$ prove -lv t/redirect.t
Must supply server and a port pairs. You supplied 198.51.100.123
```

With the incantation above, the test just hangs:

```
$ git diff
diff --git a/t/redirect.t b/t/redirect.t
index 8aa8c09b..a85a8673 100644
--- a/t/redirect.t
+++ b/t/redirect.t
@@ -2,7 +2,7 @@ use strict;
 use warnings;

 use Test::More;
-use Test::RequiresInternet;
+use Test::RequiresInternet ( '198.51.100.123' => 80 );
 use LWP::UserAgent ();

 # This is a regression test for #171
```

With the incantation in this patch:

```
$ prove -lv t/redirect.t

ok 1 - Timeout gives a client warning
ok 2 - ... and has tells us about the problem
ok 3 - Timeout with no redirects gives a client warning
ok 4 - ... and has tells us about the problem
ok
All tests successful.
Files=1, Tests=4,  2 wallclock secs ( 0.02 usr  0.00 sys +  0.09 cusr  0.02 csys =  0.13 CPU)
Result: PASS
```

```
$ NO_NETWORK_TESTING=1 prove -lv t/redirect.t
skipped: NO_NETWORK_TESTING
Files=1, Tests=0,  0 wallclock secs ( 0.01 usr  0.00 sys +  0.05 cusr  0.01 csys =  0.07 CPU)
Result: NOTEST
```

oalders: Regarding the IP, there's a comment in this commit: https://github.com/libwww-perl/libwww-perl/commit/63bbe49114eecbd57cfd89309d2cbe96fc236edb

https://tools.ietf.org/html/rfc5737

> 3.  Documentation Address Blocks
> 
>    The blocks 192.0.2.0/24 (TEST-NET-1), 198.51.100.0/24 (TEST-NET-2),
>    and 203.0.113.0/24 (TEST-NET-3) are provided for use in
>    documentation.

vanHoesel: correct me if I'm wrong: `198.51.100.123'` is not in `198.51.100.0/24`

karenetheridge: 198.51.100.0/24 should be 192.51.100.0 right through to 192.51.100.255 (/24 encompasses the entire bottom octet).

karenetheridge (approved):
LGTM!

PR #355, opened 2020-12-14
Fix RT #81381 - Make LWP::UserAgent robust to 5.17.6/5.18 hash randomization

oalders:
This is a rebased and slightly edited version of https://github.com/gisle/libwww-perl/pull/3 by @demerphq 

Audited and fixed any potential hash order dependency bugs in
LWP::UserAgent. I replaced all the uses of each() with keys() to avoid
hash iterator state bugs, and made sure that keys are sorted where
their order might matter, which was as far as I could tell only in
how proxy configuration was read from the envrionment.

There was ambiguity as to which of $ENV{http_proxy} and
$ENV{HTTP_PROXY} would be chosen by LWP::UserAgent->env_proxy(). We now
choose HTTP_PROXY if both are set, and if they differ we warn about
the conflicting configuration.

This patch includes tests to check that we warn on conflicting config
and that we correctly handle setting via either.

This includes a version bump to 6.05.

See also:

libwww-perl RT Ticket
    https://rt.cpan.org/Ticket/Display.html?id=81381

bleadperl 5.17.6 patch (queued for 5.18):
    http://perl5.git.perl.org/perl.git/commit/7dc8663964c66a698d31bbdc8e8abed69bddeec3

Eliminating the "rehash" mechanism for 5.18
    http://www.nntp.perl.org/group/perl.perl5.porters/2012/10/msg194813.html

Switch perl's hash function to MurmurHash-32 (v3) and hash randomization by default.
    http://www.nntp.perl.org/group/perl.perl5.porters/2012/11/msg195492.html

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,8 @@
 Change history for libwww-perl
 
 {{$NEXT}}
+    Fix RT #81381 - Make LWP::UserAgent robust to 5.17.6/5.18 hash
+    randomization (GH#355) (Yves Orton and Olaf Alders)
 
 6.49      2020-09-24 00:27:56Z
     - Require network testing enabled for t/redirect.t (GH#351) (Olaf Alders)
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -734,7 +734,8 @@ sub ssl_opts {
 	return $old;
     }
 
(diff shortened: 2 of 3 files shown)


simbabque (approved):
lgtm

oalders:
Thanks @simbabque!

Issue #370, opened 2021-01-11
redirect.t test failure


abraxxa:
On Debian 10 with system Perl 5.28.1:

```
PERL_DL_NONLAZY=1 "/usr/bin/perl" "-MExtUtils::Command::MM" "-MTest::Harness" "-e" "undef *Test::Harness::Switches; test_harness(0, 'blib/lib', 'blib/arch')" t/*.t t/base/*.t t/base/protocols/*.t t/leak/*.t t/local/*.t t/robot/*.t
# 
# Versions for all modules listed in MYMETA.json (including optional ones):
# 
# === Configure Requires ===
# 
#     Module                       Want     Have
#     ------------------------ -------- --------
#     CPAN::Meta::Requirements 2.120620    2.140
#     ExtUtils::MakeMaker           any     7.34
#     File::Copy                    any     2.33
#     Getopt::Long                  any      2.5
#     Module::Metadata              any 1.000033
# 
# === Build Requires ===
# 
#     Module              Want Have
#     ------------------- ---- ----
#     ExtUtils::MakeMaker  any 7.34
# 
# === Test Requires ===
# 
#     Module                 Want     Have
#     ---------------------- ---- --------
#     ExtUtils::MakeMaker     any     7.34
#     File::Spec              any     3.74
#     FindBin                 any     1.51
#     Test::Fatal             any    0.016
#     Test::More             0.96 1.302133
#     Test::Needs             any 0.002006
#     Test::RequiresInternet  any     0.05
# 
# === Test Recommends ===
# 
#     Module              Want     Have
#     --------------- -------- --------
#     CPAN::Meta      2.120900 2.150010
#     Test::LeakTrace      any  missing
# 
# === Runtime Requires ===
# 
#     Module                Want Have
#     --------------------- ---- ----
#     Digest::MD5            any 2.55
#     Encode                2.12 2.97
#     Encode::Locale         any 1.05
#     File::Listing            6 6.14
#     HTML::Entities         any 3.69
#     HTML::HeadParser       any 3.71
#     HTTP::Cookies            6 6.10
#     HTTP::Daemon             6 6.12
#     HTTP::Date               6 6.02
#     HTTP::Negotiate          6 6.01
#     HTTP::Request            6 6.18
#     HTTP::Request::Common    6 6.18
#     HTTP::Response           6 6.18
#     HTTP::Status          6.18 6.18
#     IO::Select             any 1.39
#     IO::Socket             any 1.39
[...]

PR #380, opened 2021-06-14
Attempt to avoid rare fails in redirect.t

johannessen:
See https://github.com/libwww-perl/libwww-perl/issues/370#issuecomment-859909166 for discussion.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 Change history for libwww-perl
 
 {{$NEXT}}
+    - Attempt to avoid rare fails in redirect.t (GH#380) (Arne Johannessen)
 
 6.54      2021-05-06 17:53:56Z
     - Be explicit in the prerequisite of HTTP::Status (GH#378) (Max Maischein)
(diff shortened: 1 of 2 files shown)


oalders (dismissed):
Thanks, @johannessen!

oalders:
@johannessen could you add the following to `Changes` so that the build can run?

```
$ git diff
diff --git a/Changes b/Changes
index d8a25db5..8fe56d27 100644
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@

 {{$NEXT}}
+    - Attempt to avoid rare fails in redirect.t (GH#380) (Arne Johannessen)

 6.54      2021-05-06 17:53:56Z
     - Be explicit in the prerequisite of HTTP::Status (GH#378) (Max Maischein)
 ```

johannessen:
Right, sorry about that!

oalders (approved):
👍

johannessen:
Oops … once more, now without the leading `+` from the patch.

oalders (approved):
👍

johannessen:
Thank you for your patience!

oalders:
@johannessen sorry about that! Ideally I'd just create the Changelog myself on release and you wouldn't have to bother with it. :) I've just uploaded to CPAN. Thanks for this!

PR #383, opened 2021-08-11
increase test coverage for env_proxy()

### LINKED ISSUE #372: environment case-insensitive warning painful
As of LWP::UserAgent 6.52, and perhaps somewhat earlier, it started to complain about case-insensitive "clashes" of environment variables when env_proxy was set. Even for environment variables that have nothing to do with LWP, and even when the environment is case sensitive (e.g., Unix :). This is frustrating, since it newly requires "cleaning" things which, it seems to me, LWP should not be complaining about. For example, say we have the trivial tryua.pl:
```
use LWP; my $ua = LWP::UserAgent->new( env_proxy => 1 );
```
and then run
```
env foo=1 FOO=2 perl tryua.pl
```
There is output:
```
Environment contains multiple differing definitions for 'foo'.                  
Using value from 'FOO' (2) and ignoring 'foo' (1) at /usr/local/lib/perl5/site_\
perl/5.32.1/LWP/UserAgent.pm line 1117.        
```

(Naturally, in practice the issue is not with explicitly-specified envvars like this, but case-varying envvars that I have set in my environment for reasons unrelated to LWP.)

At least for me, tt would be nice if either LWP wouldn't complain on case-sensitive systems, or wouldn't complain about envvars that it doesn't ever read, or at least had an option to suppress these checks. Thanks for considering.

eserte:
Before doing changes to fix #372 it is good to have a
complete test coverage of env_proxy().

### DIFF
--- a/Changes
+++ b/Changes
@@ -2,6 +2,7 @@ Change history for libwww-perl
 
 {{$NEXT}}
     - Update the CONTRIBUTING doc to no longer reference TravisCI. (GH #384) (Slaven Rezić)
+    - Increase test coverage for env_proxy() (GH#383) (Slaven Rezić)
 
 6.55      2021-06-17 13:57:06Z
     - Attempt to avoid rare fails in redirect.t (GH#380) (Arne Johannessen)
--- a/t/base/proxy.t
+++ b/t/base/proxy.t
@@ -4,7 +4,10 @@ use Test::More;
 use Test::Fatal;
 
 use LWP::UserAgent;
-plan tests => 2;
+plan tests => 8;
+
+# in case already defined in user's environment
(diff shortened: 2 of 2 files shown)


oalders:
Thanks @eserte! Could you add the following to the Changes file?

`    - Increase test coverage for env_proxy() (GH#383) (Slaven Rezić)`

I need to tweak the workflow not to run that test on PRs, but this will help for today.

eserte:
Unfortunately the tests on Github Actions are failing for perls <= 5.10. It seems to be an issue with Pod::Eventual, which needs a newer perl. Here's a run with perls <= 5.12 removed: https://github.com/eserte/libwww-perl/actions/runs/1120635314 (removal of 5.12 here was a mistake).

oalders:
If you could rebase to get the latest changes from `master`, the test failures should now be fixed.

oalders (approved):
Thanks @eserte!

PR #385, opened 2021-08-16
When a truthy Content-Type is provided, override the default

wolfsage:
If you set up a default_header('Content-Type => '...') but then
do ->post($url, 'Content-Type' => 'foo', Content => "..."), the
provided content type should override the default!

We only override the default if the provided value is truthy though,
as this mirrors how ->post($url, "Content-Type" => 0) behaves.

This fixes a bug in my initial implementation in #100

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -445,38 +445,40 @@ sub get {
     return $self->request( HTTP::Request::Common::GET( @parameters ), @suff );
 }
 
-sub _has_raw_content {
+sub _maybe_copy_default_content_type {
     my $self = shift;
-    shift; # drop url
+    my $req = shift;
 
-    # taken from HTTP::Request::Common::request_type_with_data
+    my $default_ct = $self->default_header('Content-Type');
+    return unless defined $default_ct;
+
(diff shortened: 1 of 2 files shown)


oalders (approved):
Thanks, @wolfsage!

genio:
@wolfsage Can you run perltidy on both `_maybe_copy_default_content_type` and the `default_content_type.t` test file, please? We have a perltidyrc in place, but this is really old code that has style problems everywhere. Our current strategy is to tidy whatever we're making changes to so that hopefully it will one day all be tidy in the same style.

`perltidy -pro=.perltidyrc lib/LWP/UserAgent.pm -o foo.pm`

That should dump the entire tidied file into `foo.pm` from which you can copy/paste the `_maybe_copy_default_content_type` function.

Same for the test file, but just go ahead and tidy that entire file.

Thank you!

oalders:
Thanks, @genio! I can go ahead and apply the tidying after merge. I'll do that now.

wolfsage:
Thanks!

oalders:
libwww-perl-6.56.tar.gz has now been uploaded.

PR #386, opened 2021-09-20
Update docs for protocols_allowed and protocols forbidden

oalders:


genio (approved):
Wording change looks good to me.

PR #388, opened 2021-10-04
improve documentation readability

simbabque:


lib/LWP/ConnCache.pm
```diff
@@ -245,7 +245,7 @@ that can be cached.  The first parameter is a short string like
 Drop connections by some criteria.  The $checker argument is a
 subroutine that is called for each connection.  If the routine returns
 a TRUE value then the connection is dropped.  The routine is called
-with ($conn, $type, $key, $deposit_time) as arguments.
+with (C<$conn>, C<$type>, C<$key>, C<$deposit_time>) as arguments.
 
 Shortcuts: If the $checker argument is absent (or C<undef>) all cached
```
vanHoesel, line 250: ☝️

simbabque: I'm not sure I understand what you mean, but I've made the whole list one code element now.

oalders (approved):
Thanks, @simbabque!

lib/LWP/ConnCache.pm
vanHoesel: it was about `$checker`

PR #389, opened 2021-10-04
Replace remaining uses of "base" with "parent"

JRaspass:
Base says 
> Unless you are using the fields pragma, consider this module discouraged in favor of the lighter-weight parent.

We aren't using fields so let's follow their advice, parent is also core in 5.10+.

We were already using parent in one place (Protocol::http) though it was undeclared in our deps so this fixes that. Depend on the minimum version to support `-norequire`.

### DIFF
--- a/META.json
+++ b/META.json
@@ -77,7 +77,7 @@
             "URI" : "1.10",
             "URI::Escape" : "0",
             "WWW::RobotRules" : "6",
-            "base" : "0",
+            "parent" : "0.217",
             "perl" : "5.008001",
             "strict" : "0",
             "warnings" : "0"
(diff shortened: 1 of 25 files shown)


META.json
```diff
@@ -77,7 +77,7 @@
             "URI" : "1.10",
             "URI::Escape" : "0",
             "WWW::RobotRules" : "6",
-            "base" : "0",
+            "parent" : "0.217",
```
karenetheridge, line 80: What's your reasoning for requiring this specific version?

JRaspass: Depend on the minimum version to support `-norequire`. As I noted in the commit and PR message.

oalders (approved):
Thanks, @JRaspass!

PR #391, opened 2021-11-15
Remove unnecessary string concatenations in mirror error messages

scop:
No longer useful after related linefeeds were removed in 21bfce4df2767929b79e15b3dc4c27db211772a4.


genio (approved):
Wow. Thanks again for the fix

PR #392, opened 2021-11-15
Fix response_done file close error message

scop:
We are no longer writing to the file at this point.


genio (commented):
I'm not so sure on this one. While I agree with the sentiment of the change, I've run into prior work where altering an error message led to downstream module tests failing because authors were relying on the text of the error message.

petdance:
I think "closing" vs "writing" in this case is a distinction without a difference.  In both cases, any I/O failure would be because of an inability to write to disk.

Issue #362, opened 2020-12-23
Unable to securely download to a tempfile [rt.cpan.org #121088]


oalders:
Migrated from [rt.cpan.org#121088](https://rt.cpan.org/Ticket/Display.html?id=121088) (status was 'new')

Requestors:
* tlhackque@yahoo.com

From tlhackque@yahoo.com on 2017-04-12 13:28:15
:
```
Background
==========
The ":content_file => $filename" header tells LWP::UserAgent to place the contents of a retrieved file into the specified filename.

However, some applications want to download to a temporary file.  This can't be done securely with the current API.  At least, not without writing handlers/callbacks.

File::Temp is the (most common) supported way to produce a safe temporary file.  It returns a file handle. (It also handles deleting the temporary file when it goes out of scope or is closed.)

One can obtain the corresponding filename and pass it to :content_file - but this introduces a race condition.  An attacker might symlink the filename to an important file, which LWP would open for write...  Also, the open() in LWP will supersede the temporary file, so the File::Temp handle is useless for reading the data.  Thus, the caller must re-open() the file by name - which is another race condition.  Finally, on some OSes, superseding an open file will fail.

Suggestion
==========
A straightforward solution would be to add a "content_fh => <filehandle>" option.  This would bypass the open() in LWP and simply write data to the specified file handle.  

In this case, LWP should not close() the file handle, or attempt to position it.  (In some other applications, the fh could be non-seekable (e.g. a pipe).)

The caller would be responsible for seek()ing to the start of data if desired (in the temp file case) and for closing the file handle.

This solution would also support passing an anonymous temporary file handle (open() of undef), which on some operating systems doesn't have a filename.

Also, the documentation could be clearer that :content_file stores decoded_content...

Thanks!


```

PR #393, opened 2021-11-15
Mirror temporary file handling fixes

scop:


genio (changes requested):
This is another one I agree with, but it has a minor gotcha:

* File::Temp is not yet a prereq. Can you please run through the build process to get that change added in as part of the PR as well.

If you want assistance with Dist::Zilla, we're happy to help.

scop:
Ah, I didn't notice core modules are listed too (can understand why one might want that though).

But I'm not having any luck with getting `dzil` to change anything in files registered in Git. Tried the commands in CONTRIBUTING.md's "Installing and using Dist::Zilla" section but they didn't do that. `dzil regenerate` didn't work after running the dep install recipes, needed to manually install `Dist::Zilla::App::Command::regenerate`, and after doing so it runs but ends up complaining "No -Regenerator plugins to regenerate with at [...]".

genio:
Ah, that's because I'm a dummy and assumed it was working the other way around like some of our dists. In this one, it pulls the prereqs _from_ the cpanfile.

Update the cpanfile to add it, do a dzil test, then dzil regenerate and all should be good.

scop:
Did the above, but `dzil regenerate` still does nothing but complains "No -Regenerator plugins to regenerate with at [...]".

Aside, if we want to use regenerate, should `Dist::Zilla::App::Command::regenerate` not be added to some dependency set?

Also, tests fail in in `en_GB.UTF-8` environment (xt/author/live/jigsaw/auth-d.t, xt/author/pod-spell.t), pass in `en_US.UTF-8`.

oalders:
Unless I'm misunderstanding something, just updating the `cpanfile` should be fine. I will update the `dzil` boilerplate on release.

Also, if we could add:

`    - Mirror temporary file handling fixes (GH#393) (Ville Skyttä` to `Changes` that would get the other tests running.

I'm not worried about the `xt` failures, unless someone wants to fix them. The jigsaw tests in particular are pretty brittle.

scop:
`Changes` and `cpanfile` updated.

skaji:
File::Temp creates files with 0600 file permissions (at least, File::Temp version 0.2311).
This may break existing behavior.

scop:
> File::Temp creates files with 0600 file permissions (at least, File::Temp version 0.2311).

Very weird, I'm *sure* I tested this case locally and found that only `mktemp` did, and that `tempfile` wouldn't. But I do see the 0600's now and agree this needs to be addressed. Thanks!

(Aside, I also think that overwriting an existing file from mirror should preserve the existing file's ownership and mode, but that's something for a different day. Perhaps I should report that as a bug so it doesn't get lost.)

lib/LWP/UserAgent.pm
```diff
@@ -1039,6 +1043,11 @@ sub mirror
             rename( $tmpfile, $file )
                 or die "Cannot rename '$tmpfile' to '$file': $!\n";
 
+            my $umask = umask();
+            my $mode = 0666 &~ ($umask || 0);
```
scop, line 1047: I'm undecided if `0666` is good when `umask` is not supported (== returns `undef`), or if we should use some other perms (`0644`?) or skip `chmod` altogether when that's the case.

lib/LWP/UserAgent.pm
```diff
@@ -1039,6 +1043,10 @@ sub mirror
             rename( $tmpfile, $file )
                 or die "Cannot rename '$tmpfile' to '$file': $!\n";
 
+            my $mode = 0666 &~ (umask() || 0);
```
scop, line 1046: I'm undecided if 0666 is good when umask is not supported (== returns undef), or if we should use some other perms (0644?) or skip chmod altogether when that's the case.

scop:
Rebased to resolve conflicts. The issue flagged by @skaji has been fixed for a while, and the only thing remaining is to decide what to do on systems that do not support `umask`, see above.

lib/LWP/UserAgent.pm
scop: Any thoughts on this?

lib/LWP/UserAgent.pm
```diff
@@ -1005,10 +1005,14 @@ sub mirror
             $request->header( 'If-Modified-Since' => HTTP::Date::time2str($mtime) );
         }
     }
-    my $tmpfile = "$file-$$";
+
+    require File::Temp;
+    my ($tmpfh, $tmpfile) = File::Temp::tempfile("$file-XXXXXX");
+    close($tmpfh) or die "Could not close tmpfile '$tmpfile': $!";
```
afresh1, line 1011: Perhaps instead of this we could teach LWP::Protocol to [dup filehandles](https://perldoc.perl.org/functions/open#Duping-filehandles):
```
diff --git a/lib/LWP/Protocol.pm b/lib/LWP/Protocol.pm
index ef12a40f..a220a487 100644
--- a/lib/LWP/Protocol.pm
+++ b/lib/LWP/Protocol.pm
@@ -110,8 +110,9 @@ sub collect
             if (!defined($arg) || !$response->is_success) {
                 $response->{default_add_content} = 1;
             }
-            elsif (!ref($arg) && length($arg)) {
-                open(my $fh, ">", $arg) or die "Can't write to '$arg': $!";
+            elsif (fileno($arg) || !ref($arg) && length($arg)) {
+                my $mode = fileno($arg) ? '>&' : '>';
+                open(my $fh, $mode, $arg) or die "Can't write to '$arg': $!";
                 binmode($fh);
                 push(@{$response->{handlers}{response_data}}, {
                     callback => sub {
```
and then this could become:
```
my $tmpfile = File::Temp->new("$file-XXXXXX");
```

And then, with the `dup` we no longer have a race condition on the tmpfile name which I think would close #362.

lib/LWP/UserAgent.pm
oalders: I wonder if skipping `chmod` would be the least confusing thing to do when `umask` is not supported.

lib/LWP/UserAgent.pm
afresh1: I filed a PR for that over in #400

lib/LWP/UserAgent.pm
scop: For better or for worse, that'll leave us with whatever was set for the tmpfile. Do we have anyone listening here who would have access to a system that doesn't support `umask`, and could confirm what do file permissions for tempfiles created the way we do here look like?

One thing we could do is to copy over the mode bits of the previous file if it existed, and not set ones to something we decide every time. That won't solve the problem discussed here, but would make sense to me. But then again, should we do this even if the previous file wasn't owned by us?

afresh1: I think for existing files, #396 is the correct solution.  Do note that [File::Temp has a `PERMS` flag](https://metacpan.org/pod/File::Temp) that seems to handle at least a few different operating systems.

scop: Oh, I had already forgotten filing #396, thanks for the reminder! I'll skip preserving existing perms for the purposes of this PR so we don't balloon it any more. 53f5c63 skips the `chmod` if `umask` is not supported.

afresh1: 0666 & umask seems like it gives the user the most control.  Should probably be documented that it does that though.

scop: I don't think that is correct use of umask. For example with a standard 0002 one, that gives `--------w-`, whereas `0666 &~ umask` as in this PR currently gives the expected `-rw-rw-r--`.

afresh1: Yes, what you have.  I meant "0666 with umask" and was not careful in my wording.

scop: I wouldn't bother documenting it though; that's the way commands that write files are expected to work. If they do something different, then _that_ I think should be documented.

oalders (approved):
Thanks, @scop!

oalders:
I fixed the merge conflict and rebased at the command line.

PR #394, opened 2021-11-16
Use American English aspell master dictionary for POD spelling tests

scop:
Locale settings affect the choice of aspell's dictionary unless one is explicitly set.

Refs https://github.com/libwww-perl/libwww-perl/pull/393#issuecomment-970301100


oalders (approved):
Thanks, @scop!

PR #397, opened 2021-11-17
Set modification time in an atomic way

### LINKED ISSUE #154: [PATCH] mirror: set modification time in an atomic way [rt.cpan.org #64147]
Migrated from [rt.cpan.org#64147](https://rt.cpan.org/Ticket/Display.html?id=64147) (status was 'new')

Requestors:
* SREZIC@cpan.org

Attachments:
* [0002-mirror-set-the-modification-time-before-renaming-the.patch](https://rt.cpan.org/Ticket/Attachment/872338/451751/0002-mirror-set-the-modification-time-before-renaming-the.patch)

From srezic@cpan.org on 2010-12-22 15:54:54:
```
mirror() should set the modification time of the mirrored file *before*
the temporary file is renamed. Patch attached.

Regards,
    Slaven
```

oalders:
Apply patch provided by @eserte in https://rt.cpan.org/Public/Bug/Display.html?id=64147

Fixes #154


simbabque (approved):

simbabque:
It turns out this is no longer needed, because d2912062002482210fff6f6ecd68c6884cc3c9c9 has already fixed the issue. I had already written tests to make codecov happy when I found out. After a short discussion with @oalders we decided to drop the original commit but keep the tests.

Issue #292, opened 2018-05-01
getstore does not work sanely with Path::Tiny objects (or any other object)


autarch:
This manifests as silently not writing the content to a file.

I think this is really part of a core bug in LWP, where `LWP::Protocol->collect` is probably dying but that die is being discarded.

It seems like `collect` should maybe check if the object in `$arg` can be stringified and then use that as a file path. And yes, that's broken and dumb, but so is Perl 5's overloading.

But it also seems like a bug that if `collect` is dying (which it has to be) that `getstore` doesn't catch that.

oalders:
Do you have a quick code sample to illustrate this?

autarch:
```
perl -e 'use Path::Tiny; use LWP::Simple qw( getstore ); my $p = path(q{./LWP::Simple}); getstore(q{https://metacpan.org/pod/LWP::Simple}, $p)'
```

genio:
```
perl -e 'use Path::Tiny; use LWP::Simple qw( getstore ); my $p = path(q{./LWP::Simple}); getstore(q{https://metacpan.org/pod/LWP::Simple}, "$p")'
```

Stringify the path. Worked here.

autarch:
Yes, I know stringification works. But it shouldn't fail _silently_ and returns a 200 when given an object like it currently does.

genio:
Gah. Apologies. I completely missed the point of the issue. Failure to read on my part.  Agreed, this shouldn't fail silently.

autarch:
Heh, no apology needed. I think I've done this once or twice (or ten thousand times).

genio:
After tracing it through (using http://checkip.amazonaws.com/ so that LWP::Protocol::https doesn't get involved), it seems you're right about where the error's occurring. LWP::Protocol->collect() fails since ```$arg``` fails the ref checks. When this happens, the ```x-died``` and ```client-aborted``` headers are set.

I'll look a bit longer to see where that error should be captured. I tend to agree with you at this point that it's a silent failure.

oalders:
If I clean up this PR would that fix the silent failure? https://github.com/libwww-perl/HTTP-Message/pull/25

autarch:
The warning in that PR is better than nothing, but I really think this needs a deeper fix. Maybe `getstore` need multiple return values or something so it can signal that the _request_ worked but something else failed. Or maybe it should throw an exception on a bad path to store in. I dunno.

oalders:
[...]

PR #400, opened 2021-12-20
Allow downloading to a filehandle

### LINKED ISSUE #362: Unable to securely download to a tempfile [rt.cpan.org #121088]

afresh1:
By checking that it "looks like a name" as well as if it is an openhandle.  If it is a filehandle, we "dup" it instead of opening directly.

This allows us to avoid some possible race conditions and should resolve #362 and be able to use the more secure idiom in #393 .  I think it would also cause the problem described in #292 to no longer error.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 Change history for libwww-perl
 
 {{$NEXT}}
+    - Allow downloading to a filehandle (GH#400) (Andrew Fresh)
 
 6.60      2021-12-17 22:29:57Z
     - Mirror temporary file handling fixes (GH#393) (Ville Skyttä)
--- a/lib/LWP/Protocol.pm
+++ b/lib/LWP/Protocol.pm
@@ -113,7 +113,7 @@ sub collect
             }
             elsif (defined(openhandle($arg)) || !ref($arg) && length($arg)) {
                 my $existing_fh = defined(openhandle($arg));
-                my $mode = $existing_fh ? '>&' : '>';
+                my $mode = $existing_fh ? '>&=' : '>';
                 open(my $fh, $mode, $arg) or die "Can't write to '$arg': $!";
                 binmode($fh);
                 push(@{$response->{handlers}{response_data}}, {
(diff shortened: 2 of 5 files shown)


afresh1:
Don't merge this yet, I had to leave for ZooLights and while there I realized I forgot to update all the things and forgot to check for `defined`.  Will get an update out tomorrow.

oalders:
Adding this to `Changes` should get the builds going:

```
    - Allow downloading to a filehandle (GH#400) (Andrew Fresh)
```

afresh1:
Rebased and added the Changes

simbabque (approved):
Very cool. I like it.

oalders:
Thanks @afresh1 and @simbabque!

PR #401, opened 2021-12-21
Use File::Copy::move to attempt an atomic mirror

afresh1:
In most situations this will Just Work, but some operating systems don't support an atomic rename over an existing file and there's nothing we can do about it.

In any case, the conversation around #362 was a bit overwhelming when IMO File::Copy::move already does the right thing.  I think this will fix things while documenting just enough that people will be aware of the caveats.


scop (dismissed):
Nice, LGTM :+1:

afresh1:
Do I need to do something for that?

oalders:
> Do I need to do something for that failed check?

You can add this to `Changes`:

```
    - Use File::Copy::move to attempt an atomic mirror (GH#401) (Andrew Fresh)
```

afresh1:
Rebased and added the Changes

oalders (approved):
👍

PR #406, opened 2022-01-21
recreate pr 301

### LINKED ISSUE #186: Run tests against localhost instead of hostname [rt.cpan.org #94959]

oalders:
- Issue 186: use localhost instead of IPv4 address
- Update Changes

Fixes #186

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 Change history for libwww-perl
 
 {{$NEXT}}
+    - Change the hostname to use localhost in the tests (GH#301) (mschae94)
 
 6.67      2022-06-14 20:20:12Z
     - Remove undocumented LWP::Version sub (GH#416) (James Raspass)
--- a/t/local/http.t
+++ b/t/local/http.t
@@ -679,7 +679,7 @@ sub daemonize {
 
     # Note: timeout of 0 is not infinite, so no point in special casing
     # timeout logic.
-    my $d = HTTP::Daemon->new(Timeout => $ENV{PERL_LWP_ENV_HTTP_TEST_SERVER_TIMEOUT} || 10, LocalAddr => '127.0.0.1') || die $!;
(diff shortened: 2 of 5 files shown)


oalders:
This breaks testing on macOS and Windows.

simbabque (dismissed):

simbabque:
It looks alright. Can we rerun the test suite?

oalders:
I just rebased and force-pushed. That should start a new build.

oalders:
It looks like it's `t/local/http.t` that is causing the problems.

simbabque:
I suspect it's the `$base->host` that is not matching `localhost`, but I haven't tested it yet.

simbabque:
It works on my machine. I haven't figured out yet why it fails in the GH action.

Issue #239, opened 2017-03-31
add option to automatically handle meta redirects [rt.cpan.org #95988]


oalders:
Migrated from [rt.cpan.org#95988](https://rt.cpan.org/Ticket/Display.html?id=95988) (status was 'open')

Requestors:
* pagenyon@gmail.com

From pagenyon@gmail.com on 2014-05-26 22:21:39:
```
LWP already parses the head tag by default, and makes any contained meta http-equiv statements available as response headers.
I suggest adding an option (like meta_redirect => 1) that would tell LWP to automatically follow these redirects.

A web search shows this functionality has been asked for many times over the years, though it looks like no official feature request was made in RT or github. If there is a reason this will not be implemented, can you at least include an example in the docs on how the user can implement using request handlers.
```

From pagenyon@gmail.com on 2014-05-26 23:02:47:
```
On Mon May 26 18:21:39 2014, pagenyon wrote:
> LWP already parses the head tag by default, and makes any contained
> meta http-equiv statements available as response headers.
> I suggest adding an option (like meta_redirect => 1) that would tell
> LWP to automatically follow these redirects.
> 
> A web search shows this functionality has been asked for many times
> over the years, though it looks like no official feature request was
> made in RT or github. If there is a reason this will not be
> implemented, can you at least include an example in the docs on how
> the user can implement using request handlers.

figured out how to do with request handlers:

    $ua->set_my_handler(response_done => sub{
        my ($response, $ua, $handler) = @_;
        if (my ($refresh) = $res->remove_header('Refresh')) {
            $refresh =~ s/.*?\burl=//;
            my $uri = URI->new($refresh)->canonical;
            return if ref $uri eq 'URI::_generic';
            $response->code(301);
            $response->header(Location => $uri);
        }
    }, m_code => 200);
```

From ether@cpan.org on 2014-05-26 23:15:53:
```
On Mon, May 26, 2014 at 07:02:48PM -0400, pagenyon via RT wrote:
>             return if ref $uri eq 'URI::_generic';
>             $response->code(301);
>             $response->header(Location => $uri);

Checking for types like this is fragile -- the underscore is a hint that
you shouldn't be depending on this name.  Instead, check for the
capabilities that you need - e.g.  return unless $uri->can('scheme'),
or return unless $uri->some_property eq $value.
```

grr:
is this something that would accepted if somebody implemented a better solution? i have something similar in my toolbox that could be adapted.

[...]

Issue #196, opened 2017-03-31
parse_head fails if content is encoded [rt.cpan.org #107332]


oalders:
Migrated from [rt.cpan.org#107332](https://rt.cpan.org/Ticket/Display.html?id=107332) (status was 'new')

Requestors:
* justincase@yopmail.com

From justincase@yopmail.com on 2015-09-25 20:24:11:
```
I was trying to implement meta-equiv-refresh handling as described in this ticket, https://rt.cpan.org/Public/Bug/Display.html?id=95988

But the refresh header was missing from my responses. This was because I was setting this header "accept_encoding => 'gzip,deflate'", which resulted in an encoded response, thus HTML::HeadParser failed to parse it as html. Can we have an option to automatically decode encoded content if the content_type is 'text/html'?
```

PR #409, opened 2022-04-05
meta refresh redirect

grr:
Satisfies #239

I noticed two existing issues with the headparser code, neither of which I have addressed in this commit:

1. https://github.com/libwww-perl/libwww-perl/issues/196)
2. Noscript tags are valid within the head tag, and they can contain meta tags. So it's possible to have two sets of contradictory meta tags. We can either:
    * ignore noscript: `$parser->ignore_elements('noscript');`
    * or distinguish the meta tags found in noscript by prefixing/suffixing them so the user can choose which to use.

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -751,26 +751,43 @@ sub parse_head {
         my $flag = shift;
         my $parser;
         my $old = $self->set_my_handler("response_header", $flag ? sub {
-               my($response, $ua) = @_;
-               require HTML::HeadParser;
-               $parser = HTML::HeadParser->new;
-               $parser->xml_mode(1) if $response->content_is_xhtml;
-               $parser->utf8_mode(1) if $] >= 5.008 && $HTML::Parser::VERSION >= 3.40;
+                my($response, $ua) = @_;
+                require HTML::HeadParser;
+                $parser = HTML::HeadParser->new;
+                $parser->xml_mode(1) if $response->content_is_xhtml;
+                $parser->utf8_mode(1) if $] >= 5.008 && $HTML::Parser::VERSION >= 3.40;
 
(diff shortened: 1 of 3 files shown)


lib/LWP/UserAgent.pm
```diff
@@ -751,26 +751,43 @@ sub parse_head {
...
+                            my $h = $parser->header;
+                            my $r = $_[0];
+                            for my $f ($h->header_field_names) {
+                                $r->init_header($f, [$h->header($f)]);
+                            }
+                            undef($parser);
+                        }
+                    },
+                });
+                push(@{$response->{handlers}{response_redirect}}, {
```
oalders, line 773: I wonder if this should check `max_redirect`? If, for instance, `max_redirect` is 0, it might be surprising for this to happen.

oalders:
@grr could you apply the following and push, so that we can get past the build test failure?

```diff
diff --git a/Changes b/Changes
index 3a5bf9f7..0fb4c209 100644
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 Change history for libwww-perl

 {{$NEXT}}
+   - Redirect on meta refresh (GH#409) (grr)

 6.62      2022-04-05 01:04:17Z
     - Allow downloading to a filehandle (GH#400) (Andrew Fresh)
```

simbabque (approved):
lgtm

oalders:
Thanks @grr and @simbabque!

oalders:
This has now been released to the CPAN.

oalders:
@grr this had to be rolled back in #415 for reasons detailed there. We could take another pass at this if you like. Basically the `WWW::Mechanize::Cached` issue would need to be addressed and we'd need to disable the behaviour by default so that it's less surprising.

PR #415, opened 2022-05-17
Revert automatic follow of meta refresh tags

oalders:
- Revert "Remove unused variable introduced in 9d73bc422"
- Revert "Add tests for meta_refresh + max_redirects"
- Revert "Preserve the Refresh header on meta redirect"
- Revert "Add tests for meta refresh redirects"
- Revert "Handle meta refresh redirects"

This also broke WWW::Mechanize::Cached: https://github.com/libwww-perl/WWW-Mechanize-Cached/issues/27

From #toolchain on irc.perl.org

```
Could I get some feedback on https://github.com/libwww-perl/WWW-Mechanize-Cached/pull/28 Specifically https://github.com/libwww-perl/WWW-Mechanize-Cached/pull/28/commits/ab875678759d66300d772e33e19d76d143970ed4
I was getting "Can't store CODE items" errors from Storable when caching HTTP::Response objects.

feels wrong
LWP is leaving a handler in place unexpectedly, when it removes others
also, the PR that added the handler feels wrong
having LWP treat a Refresh header as a redirect seems very wrong

it _might_ be defensible if the delay is specified as 0

yeah, Refresh != redirect
```

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,8 @@
 Change history for libwww-perl
 
 {{$NEXT}}
+    - Revert automatic follow of meta refresh tags which was added in 6.63
+      (GH#415) (Olaf Alders)
 
 6.65      2022-05-09 18:36:14Z
     - fix NAME in Makefile.PL (GH#413) (Graham Knop)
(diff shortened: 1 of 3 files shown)


haarg (approved):
This looks reasonable to me. It does seem like a feature that could be supported, but not enabled by default.

The issue with leaving the handler in place and causing cloning problems is also something that should be followed up on.

PR #418, opened 2022-10-03
Apply perlimports to tests

oalders:
- Add perlimports.toml
- Apply perlimports to tests

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 Change history for libwww-perl
 
 {{$NEXT}}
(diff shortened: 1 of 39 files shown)


simbabque (approved):
LGTM

simbabque (approved):
Great, thank you.

simbabque:
I think we can ignore the coverage. If it's accurate I have no idea how it lost partial coverage of that line by moving around imports in tests. Makes me wonder how trustworthy that report really is.

PR #419, opened 2022-10-03
remove dynamic dependency on HTTP::Status

haarg:
The dynamic dependency on HTTP::Status was meant to avoid one known problematic version. But upgrading HTTP::Status is not an onerous requirement. Simplify the config and dependencies by just requiring the new working version.


oalders (approved):
👍🏻

Issue #226, opened 2017-03-31
Incorrect encoding handling for text/html files with LWP::Simple::get [rt.cpan.org #69393]


oalders:
Migrated from [rt.cpan.org#69393](https://rt.cpan.org/Ticket/Display.html?id=69393) (status was 'open')

Requestors: @vinc17fr on 2011-07-11

genio:
@vinc17fr wrote:

When a file declared as iso-8859-1 and served as text/html is also a
valid UTF-8 file, LWP::Simple::get regards it as a UTF-8 file. This is
incorrect. Unless the website from which I fetch data recently changed,
it seems to be a regression from libwww-perl 5.x.

For instance, with lwp-dump being

```perl
#!/usr/bin/env perl

use strict;
use Devel::Peek;
use LWP::Simple;

@ARGV == 1 or die "Usage: $0 <URL>\n";
my $url = shift;
my $file = LWP::Simple::get($url);
defined $file or die "$0: can't fetch $url\n";
Dump $file;
```

and when running

```bash
for i in 1a 1h 2a 2h
do
  ./lwp-dump http://www.vinc17.net/test/perl-lwp-test$i.xml \
      2> perl-lwp-test$i.dump
done
```

I get:

```
==> perl-lwp-test1a.dump <==
SV = PV(0x194dac8) at 0x6a02d0
  REFCNT = 1
  FLAGS = (PADMY,POK,pPOK,UTF8)
  PV = 0x1308cd0 "<?xml version=\"1.0\"
encoding=\"iso-8859-1\"?>\n<root>post\303\203\302\251... A</root>\n"\0
[UTF8 "<?xml version="1.0"
encoding="iso-8859-1"?>\n<root>post\x{c3}\x{a9}... A</root>\n"]
  CUR = 71
  LEN = 80

==> perl-lwp-test1h.dump <==
SV = PV(0x194dac8) at 0x6a02d0
  REFCNT = 1
  FLAGS = (PADMY,POK,pPOK,UTF8)
  PV = 0x13097d0 "<?xml version=\"1.0\"
encoding=\"iso-8859-1\"?>\n<root>post\303\251... A</root>\n"\0 [UTF8
"<?xml version="1.0" encoding="iso-8859-1"?>\n<root>post\x{e9}...
A</root>\n"]
  CUR = 69
  LEN = 80

==> perl-lwp-test2a.dump <==
SV = PV(0x194dac8) at 0x6a02d0
[...]

PR #420, opened 2022-10-17
Fix `lwp-request` (`/usr/bin/GET`) for files non-ASCII

ctrlcctrlv:
Closes libwww-perl/libwww-perl#226. _I think._ This is the first time I've ever tried to patch any FOSS related to Perl.

### DIFF
--- a/t/base/protocols/file.t
+++ b/t/base/protocols/file.t
@@ -1,4 +1,5 @@
 #!/usr/bin/perl
+use utf8;
 use strict;
 use warnings;
 use Test::More tests => 4;
--- a/t/base/protocols/file.t
+++ b/t/base/protocols/file.t
@@ -1,0 +1,33 @@
+#!/usr/bin/perl
+use strict;
+use warnings;
+use Test::More tests => 4;
+
+use File::Basename qw(dirname);
+use LWP::Simple qw(get);
+use File::Temp qw(tempfile);
+
+require POSIX;
+POSIX::setlocale(&POSIX::LC_ALL, 'ja_JP.UTF-8');
+
+my $tmp, my $tmpd;
+$tmpd = File::Temp->newdir();
+$tmpd->unlink_on_destroy(1);
+$tmp = File::Temp->new(TEMPLATE => 'wwwdata_tempXXXX',
+                       DIR => $tmpd,
+                       SUFFIX => '.コピペ.test');
+
+open(TMP, '>:utf8', $tmp);
+print TMP 'テスト';
+close(TMP);
+
+## Test default directory output.
+is($tmp->filename =~ /コピペ/, 1);
+my $res = get('file://' . dirname($tmp->filename));
+is(($res =~ m/コピペ/, $&), 'コピペ');
+# Make sure there are no HTML entities.
+isnt(($res =~ m/&/, $&), '&');
+
+## Test file output.
+my $res2 = get('file://' . $tmp->filename);
+is($res2, 'テスト');
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 Change history for libwww-perl
 
 {{$NEXT}}
+    - Fix `lwp-request` (`/usr/bin/GET`) for files non-ASCII (Fredrick Brennan)
 
 6.68      2023-02-27 19:18:33Z
     - Remove dynamic dependency on HTTP::Status (GH#419) (Graham Knop)
--- a/bin/lwp-request
+++ b/bin/lwp-request
@@ -182,6 +182,9 @@ Gisle Aas <gisle@aas.no>
 
 use strict;
 use warnings;
+use locale;
+require POSIX;
+POSIX::setlocale(&POSIX::LC_ALL, "") and POSIX::setlocale(&POSIX::LC_CTYPE, "");
 
 my $progname = $0;
 $progname =~ s,.*[\\/],,;    # use basename only
--- a/lib/LWP/Protocol/file.pm
+++ b/lib/LWP/Protocol/file.pm
@@ -90,6 +90,8 @@ sub request
 	# Make directory listing
 	require URI::Escape;
 	require HTML::Entities;
+	require POSIX;
+	my $localeenc = POSIX::setlocale(&POSIX::LC_CTYPE) =~ s/.*\.//gr;
         my $pathe = $path . ( $^O eq 'MacOS' ? ':' : '/');
 	for (@files) {
 	    my $furl = URI::Escape::uri_escape($_);
@@ -97,7 +99,7 @@ sub request
                 $furl .= '/';
                 $_ .= '/';
             }
-	    my $desc = HTML::Entities::encode($_);
+	    my $desc = HTML::Entities::encode($_, '<>&"');
 	    $_ = qq{<LI><A HREF="$furl">$desc</A>};
 	}
 	# Ensure that the base URL is "/" terminated
@@ -109,6 +111,7 @@ sub request
 			"<HTML>\n<HEAD>",
 			"<TITLE>Directory $path</TITLE>",
 			"<BASE HREF=\"$base\">",
+			"<META CHARSET=\"$localeenc\">",
 			"</HEAD>\n<BODY>",
 			"<H1>Directory listing of $path</H1>",
 			"<UL>", @files, "</UL>",
--- a/lib/LWP/Protocol/file.pm
+++ b/lib/LWP/Protocol/file.pm
@@ -11,6 +11,7 @@ require HTTP::Request;
 require HTTP::Response;
 require HTTP::Status;
 require HTTP::Date;
+require Encode::Locale;
 
 
 sub request
@@ -91,8 +92,8 @@ sub request
 	require URI::Escape;
 	require HTML::Entities;
 	require POSIX;
-	my $localeenc = POSIX::setlocale(&POSIX::LC_CTYPE) =~ s/.*\.//gr;
-        my $pathe = $path . ( $^O eq 'MacOS' ? ':' : '/');
+	my $localeenc = $Encode::Locale::ENCODING_LOCALE_FS;
+	my $pathe = $path . ( $^O eq 'MacOS' ? ':' : '/');
 	for (@files) {
 	    my $furl = URI::Escape::uri_escape($_);
             if ( -d "$pathe$_" ) {


ctrlcctrlv:
Note to maintainers: take care when merging as [e58f79a](https://github.com/libwww-perl/libwww-perl/pull/420/commits/e58f79ad653e73e10f6920818b16210d18945ca6) bumps the version of Perl required.

> ### Bump required Perl version to 5.013002
>
> I don't even know if I'm allowed to do this but that's a super old
version, between 12 April 2010 and 14 May 2011.

karenetheridge:
Why exactly is this new perl version required?

I suspect this might be too high at present.

ctrlcctrlv:
@karenetheridge I honestly couldn't tell you. Tests fail without it. Looking at the changelog, a lot of Unicode changes occurred around then, but that's honestly conjecture. I don't know enough about Perl development to say.

ctrlcctrlv:
Also, I noticed there was no test file for `LWP::Protocol::file` so I tried to make the test file I just added in bfd6f6b generic.

```perl
#!/usr/bin/perl
use strict;
use warnings;
use Test::More tests => 4;

use File::Basename qw(dirname);
use LWP::Simple qw(get);
use File::Temp qw(tempfile);

require POSIX;
POSIX::setlocale(&POSIX::LC_ALL, 'ja_JP.UTF-8');

my $tmp, my $tmpd;
$tmpd = File::Temp->newdir();
$tmpd->unlink_on_destroy(1);
$tmp = File::Temp->new(TEMPLATE => 'wwwdata_tempXXXX',
                       DIR => File::Temp->newdir(),
                       SUFFIX => '.コピペ.test');

open(TMP, '>:utf8', $tmp);
print TMP 'テスト';
close(TMP);

## Test default directory output.
is($tmp->filename =~ /コピペ/, 1);
my $res = get('file://' . dirname($tmp->filename));
is(($res =~ m/コピペ/, $&), 'コピペ');
# Make sure there are no HTML entities.
isnt(($res =~ m/&/, $&), '&');

## Test file output.
my $res2 = get('file://' . $tmp->filename);
is($res2, 'テスト');
```

haarg:
Removing the use of s///r should allow this to work on older perl versions.

I'm not sure about using the locale name as the basis for the encoding. Seems like it would need to use something like CODESET from I18N::Langinfo. Annoyingly, the `encoding` module has an internal routine to get the encoding from the locale, but won't let anyone use it.

It doesn't seem like this really resolves #226. That issue doesn't relate to `file://` URLs. Or at least, not only to `file://` URLs.

haarg:
Actually, we should be using Encode::Locale, which gives the current locale encoding as `$Encode::Locale::ENCODING_LOCALE`.

ctrlcctrlv:
Thanks @haarg! I force pushed, removing the bumped dependency, and using Encode::Locale.

ctrlcctrlv:
Whatever happened to this?

simbabque:
> Whatever happened to this?

We're all volunteers, and life got in the way, I suspect. I didn't look at github at all for the last three months, only got back round to it. Thank you for your patience.

Thanks for making the changes @haarg  suggested. I'm going to refamiliarise myself with this and have a look.

simbabque:
Closing and reopening to trigger a test run.

ctrlcctrlv:
Error: Wide character in print at ... ?

It's intentional.

oalders:
Just failing on missing Changes entry.

simbabque:
@ctrlcctrlv could you add a Changes entry and possibly rebase and force-push your changes please? A few of us are currently sitting in the same room working on OSS, and I now have time getting this pushed through. I am sorry about the wait. Thank you.

ctrlcctrlv:
Think I did that right @simbabque. I am sorry for the delay. This is my first Perl PR and I had to learn about `dzil`/`cpm` etc. to get it to break in the same way locally. Then figure out `Changes` file format. We all have to start somewhere, right :-)

t/base/protocols/file.t
```diff
@@ -0,0 +1,34 @@
...
+                       SUFFIX => '.コピペ.test');
+
+open(TMP, '>:utf8', $tmp);
+print TMP 'テスト';
+close(TMP);
+
+## Test default directory output.
+is($tmp->filename =~ /コピペ/, 1);
+my $res = get('file://' . dirname($tmp->filename));
+is(($res =~ m/コピペ/, $&), 'コピペ');
```
simbabque, line 28: I'm reading this as "the result of that match is the stuff that was matched" and the name of the test is "コピペ". Is that what you intended?

t/base/protocols/file.t
```diff
@@ -0,0 +1,34 @@
...
+# Make sure there are no HTML entities.
+isnt(($res =~ m/&/, $&), '&');
```
simbabque, line 30: Same as above. Did you mean "this should not contain an ampersand"? I'd write that as `unlike($res, qr/&/, 'output has no HTML entities')`. I suspect the bit you don't know about here is the `qr//`, which constructs a regular expression without applying it. That way you can pass a pattern around, and the `like` and `unlike` will apply it later. Here's a link to the doc: https://metacpan.org/pod/Test::More#like

simbabque (changes requested):
No worries about the delay. I was so busy with getting the tests to actually run, I didn't even look at the code myself. I left some comments about the style. I suspect you are not familiar with Perl testing. 

The new line in Changes is great, thank you for that. We require that in PRs because it makes sure we don't forget to include any changes, and makes releasing faster.

You could run the tests without dzil. It's handy to install dependencies, but a simple `prove -lv t/` from the root directory will run all the tests you need to care about when writing code. The author tests are nice, but as a drive-by contributor we don't require you to run them. That's what the github actions are for :)

simbabque:
I think the tests failed because of an issue with metacpan yesterday where dependencies could not be installed. They will rerun automatically when you push a new version.

oalders:
Close and re-open to run CI.

oalders:
`t/base/protocols/file.t` is failing under Perl 5.38

ctrlcctrlv:
i just don't understand how windows is so broken this is happening

> ``# <META CHARSET="ANSI_X3.4-1968">``

oalders:
I don't have a Windows machine. Is there anything in particular you need here?

ctrlcctrlv:
neither do i

oalders:
The failure is on Linux. Not sure what you mean about Windows.

ctrlcctrlv:
why is setlocale ignored ?

t/base/protocols/file.t
```diff
@@ -0,0 +1,36 @@
...
+use strict;
+use warnings;
+use Test::More tests => 4;
+
+use File::Basename qw(dirname);
+use LWP::Simple qw(get);
+use File::Temp qw(tempfile);
+
+require POSIX;
+POSIX::setlocale(&POSIX::LC_ALL, 'ja-JP.UTF-8');
```
oalders, line 12: I asked on irc.perl.org about `setlocale` and the feedback I got was:

> on linux, if the given locale isn't available it won't work

Having said that, this test passes locally on my Mac with those lines removed.

oalders: Also:

> also it should return the new locale if properly set
11:01 AM 
so possibly it should skip if setting the locale fails
11:02 AM 
but also, for the purposes of that test, the collation doesn't matter, so it may be better to use a locale that is more common, like en_US.UTF-8
11:02 AM 
the language part is irrelevant
11:02 AM 
as long as it exists

simbabque:
Closing and reopening to trigger another test run.

simbabque:
I've tried changing it to `en_US`, but that doesn't fix it. It returns a true value because that locale exists, but the error still occurs on my machine. I have to admit I am in over my head here.

ctrlcctrlv:
And, years later, my tests still pass.

```
[DZ] beginning to build libwww-perl
[DZ] guessing dist's main_module is lib/LWP.pm
[Git::Contributors] multiple names with the same email found: you may want to use a .mailmap file (https://www.kernel.org/pub/software/scm/git/docs/git-shortlog.html#_mapping_authors): 
[Git::Contributors]   simbabque@cpan.org
[Git::Contributors]   copypaste@kittens.ph
[Git::Contributors] multiple emails with the same name found: you may want to use a .mailmap file (https://www.kernel.org/pub/software/scm/git/docs/git-shortlog.html#_mapping_authors): 
[Git::Contributors]   slaven rezic
[GithubMeta] No user could be discerned from URL: 'https://github.com/ctrlcctrlv/libwww-perl/'
[GithubMeta] No repository could be discerned from URL: 'https://github.com/ctrlcctrlv/libwww-perl/'
[GithubMeta] skipping meta.resources.repository creation
[Prereqs::FromCPANfile] Parsing 'cpanfile' to extract prereqs
[PkgVersion, via RewriteVersion::Transitional] skipping private package RequestAgent in bin/lwp-request
[Git::Contributors] Warning: distribution has non-ascii characters in contributor names. META.json will be unparsable on perls <= 5.8.6 when JSON::PP is lower than 2.27300
[DZ] writing libwww-perl in build_dir
[CheckChangeLog] [CheckChangeLog] OK with Changes
[DZ] built in build_dir
Unknown option: g
Unknown option: show-build-log-on-failure
install is up to date. (0.01)
t/10-attrs.t ................... ok                                     
===(      10;0  1/1  0/? )==============================================# 
# Versions for all modules listed in static metadata (including optional ones):
# 
# === Configure Requires ===
# 
#     Module              Want Have
#     ------------------- ---- ----
#     ExtUtils::MakeMaker  any 7.70
#     File::Copy           any 2.41
#     Getopt::Long         any 2.57
# 
# === Configure Suggests ===
# 
#     Module      Want Have
#     -------- ------- ----
#     JSON::PP 2.27300 4.16
# 
# === Test Requires ===
# 
#     Module                 Want     Have
#     ---------------------- ---- --------
#     ExtUtils::MakeMaker     any     7.70
#     File::Spec              any     3.91
#     FindBin                 any     1.54
#     HTTP::Daemon           6.12     6.16
#     Test::Fatal             any    0.017
#     Test::More             0.96 1.302199
#     Test::Needs             any 0.002010
#     Test::RequiresInternet  any     0.05
# 
# === Test Recommends ===
# 
#     Module              Want     Have
#     --------------- -------- --------
#     CPAN::Meta      2.120900 2.150010
#     Test::LeakTrace      any  missing
# 
# === Runtime Requires ===
# 
#     Module                    Want     Have
#     --------------------- -------- --------
#     Digest::MD5                any  2.58_01
#     Encode                    2.12     3.21
#     Encode::Locale             any     1.05
#     File::Copy                 any     2.41
#     File::Listing                6     6.16
#     File::Temp                 any   0.2311
#     Getopt::Long               any     2.57
#     HTML::Entities             any     3.83
#     HTML::HeadParser          3.71     3.83
#     HTTP::Cookies                6     6.11
#     HTTP::Date                   6     6.06
#     HTTP::Negotiate              6     6.01
#     HTTP::Request             6.18     6.47
#     HTTP::Request::Common     6.18     6.47
#     HTTP::Response            6.18     6.47
#     HTTP::Status              6.18     6.47
#     IO::Select                 any     1.55
#     IO::Socket                 any     1.55
#     LWP::MediaTypes              6     6.04
#     MIME::Base64               2.1  3.16_01
#     Net::FTP                  2.58     3.15
#     Net::HTTP                 6.18     6.23
#     Scalar::Util               any     1.63
#     Try::Tiny                  any     0.32
#     URI                       1.10     5.31
#     URI::Escape                any     5.31
#     WWW::RobotRules              6     6.02
#     parent                   0.217    0.241
#     perl                  5.008001 5.040002
#     strict                     any     1.13
#     warnings                   any     1.70
# 
# === Runtime Suggests ===
# 
#     Module               Want    Have
#     -------------------- ---- -------
#     Authen::NTLM         1.02 missing
#     Data::Dump           1.13    1.25
#     IO::Socket::INET      any    1.55
#     LWP::Protocol::https 6.02    6.14
# 
t/00-report-prereqs.t .......... ok                                     
t/base/default_content_type.t .. ok                                     
t/base/protocols.t ............. ok                                     
t/base/protocols/file.t ........ ok                                     
t/base/protocols/nntp.t ........ skipped: nntp.perl.org is unstable and Test::RequiresInternet is not catching it
t/base/simple.t ................ ok                                     
t/base/proxy.t ................. ok                                     
t/base/ua_handlers.t ........... ok                                     
t/base/ua.t .................... ok                                     
t/local/autoload-get.t ......... ok                                     
t/local/autoload.t ............. ok                                     
t/leak/no_leak.t ............... skipped: Need Test::LeakTrace          
t/local/download_to_fh.t ....... ok                                     
t/local/get.t .................. ok                                     
t/local/httpsub.t .............. ok                                     
t/local/protosub.t ............. ok                                     
t/redirect.t ................... ok                                     
t/local/http.t ................. ok                                     
t/robot/ua-get.t ............... ok                                     
t/robot/ua.t ................... ok     
All tests successful.
Files=21, Tests=297,  6 wallclock secs ( 0.04 usr  0.01 sys +  1.08 cusr  0.28 csys =  1.41 CPU)
Result: PASS
[fred@大.狸.agency ~/Workspace/libwww-perl]$
```

I like Perl, the language, but I don't try to contribute anymore as it seems all the packages are abandoned and most of the language's fans retired in 2010.

simbabque:
Thank you for coming back to this. I'm going to have another look today. I think the availability of locales matters, which is why the tests break for us. 

Perl is still very much going. I'm currently in a room full of contributors at the Perl Toolchain Summit, a yearly event where we work on the health of the ecosystem around the language. There are fewer people using the language these days, of course, as it's not the first choice (and probably neither the second) for new commercial projects. But there is still a large, commercial interest in it.

PR #425, opened 2023-04-28
Recreate PR 62

simbabque:
I've rebased #62 and removed the bits about `eval` that have been superseded by `try` in the meantime.

I tried writing a test for this, but I don't fully understand how to do that. I tried adding a custom schema for which the package doesn't exist, and that lives. I could call the `implementor` function several times, but I am still unsure how to track it doesn't go looking twice if the package does not exist. But I believe we're OK here without adding a complicated test for what's fundamentally basic caching.

Closes #62.


simbabque:
I accidentally pushed this to the wrong remote :facepalm:

oalders (approved):

PR #426, opened 2023-04-29
don't allow partial matches for proxy domain names #421

### LINKED ISSUE #421: Unexpected matching mode of "no_proxy" values
Currently the `no_proxy` values are matched with a prefix match. This results in the situation that common-suffix domains may be accessed in a way not intended.

The problem seems to be in this line: https://github.com/libwww-perl/libwww-perl/blob/935e592fdad1969d0ca1fba5305784081fe0b489/lib/LWP/UserAgent.pm#L1076

This tests if a domain ends with a item in the list. I guess this is meant to allow for subdomains. However it also allows for other same-suffix domains to match.

Here is an example showing the behaviour:
```perl
use LWP::UserAgent;
my $ua = LWP::UserAgent->new;
$ua->proxy(['http'] => 'http://localhost/');
$ua->no_proxy('ample.org');
$ua->get('http://example.org/'); # connection is *not* made via proxy
```

A solution would be to do a combination of a full match and a match with a dot prefixed to the `$domain` such as `/(?:^|\.)\Q$domain\E$/i` (also notice the `i`-modifier).
Another solution would be to offer support for origin matching. See also https://github.com/libwww-perl/URI/issues/121 regarding that.

simbabque:
Fixes #421

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -1092,7 +1092,7 @@ sub _need_proxy {
     if ($ua->{no_proxy}) {
         if (my $host = eval { $req->uri->host }) {
             for my $domain (@{$ua->{no_proxy}}) {
-                if ($host =~ /\Q$domain\E$/) {
+                if ($host =~ /(?:^|\.)\Q$domain\E$/) {
                     return;
                 }
             }
--- a/Changes
+++ b/Changes
@@ -4,6 +4,8 @@ Change history for libwww-perl
     - Timeouts for cached connections now update (GH#73) (Eric Johnson)
     - The conn_cache() can now be unset (GH#424) (Julien Fiegehenn)
     - LWP::Protocol now only attempts to load modules once (GH#62) (Burak Gursoy)
+    - Fix a bug in no_proxy that allowed partial matches to a proxy address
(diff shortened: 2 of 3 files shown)


lib/LWP/UserAgent.pm
```diff
@@ -1092,7 +1092,7 @@ sub _need_proxy {
     if ($ua->{no_proxy}) {
         if (my $host = eval { $req->uri->host }) {
             for my $domain (@{$ua->{no_proxy}}) {
-                if ($host =~ /\Q$domain\E$/) {
+                if ($host =~ /\b\Q$domain\E$/) {
```
haarg, line 1095: As discussed, this should be `/(?:^|\.)\Q$domain\E$/`

oalders (approved):

defc0n:
@simbabque @oalders @haarg this breaks the ability to specify a noproxy of `.example.com` and have it match `www.example.com`. Note that this is the behavior of tools such as `curl`. I understand how `ample.com` shouldn't match `www.example.com`, but `.example.com` definitely should.

oalders:
@defc0n do you want to open an issue for this so that we can track the bug fix there?

PR #436, opened 2023-07-02
Don't mangle protocol scheme and don't require it to be valid if implementor is already known

### LINKED ISSUE #431: LWP::Protocol::socks fails since LWP 6.69
Hi!

The code below gave me “Error GETing ..: Protocol scheme 'http::socks' is not supported at ... line ...” suddenly. Turns out it happens since LWP 6.69 while it runs smoothly with LWP 6.68.

    #!/usr/local/bin/perl
    use strict;
    use LWP::UserAgent;
    my $ua = new LWP::UserAgent(agent => 'Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.8.0.5) Gecko/20060719 Firefox/1.5.0.5');
    $ua->proxy([qw(http https)] => 'socks4://localhost:9050');
    my $response = $ua->get("http://www.freebsd.org");
    print $response->code,' ', $response->message,"\n";
    my $response = $ua->get("https://www.microsoft.com");
    print $response->code,' ', $response->message,"\n";

mwgamera:
This makes LWP::Protocol::socks work and
Fixes #431


oalders:
@mwgamera could you add:

```text
    - Don't mangle protocol scheme and don't require it to be valid if implementor is already known (GH#436) (mwgamera)
```

to `Changes` and force-push? That should fix the build.

oalders:
```
Bareword found where operator expected at /home/runner/work/libwww-perl/libwww-perl/lib/LWP/Protocol.pm line 62, near "tr/.+-/_/r"
syntax error at /home/runner/work/libwww-perl/libwww-perl/lib/LWP/Protocol.pm line 62, near "tr/.+-/_/r"
```

I think `s` might be clearer that `tr` here, since it's a replacement rather than a translation. Not sure this will fix the older Perls, though.

oalders (approved):
Thanks, @mwgamera!

oalders:
This has been released to CPAN.

PR #437, opened 2023-07-17
Update CI settings

oalders:
- Run CI on 5.36 and 5.38
- Only run CI on push when branch is master
- Remove scheduled CI as it can disable the action after inactivity
- Add dependabot to upgrade actions


skaji (approved):
LGTM

oalders:
Thanks, @skaji!

PR #443, opened 2023-09-27
Allow underscores in headers not to be removed in lwp-request

### LINKED ISSUE #442: lwp-request doesn't allow non-canonicalised header names
in HTTP::Headers the headers are "canonicalized", underscores are replaced by hyphens, initial letters are uppercased.

However underscore is a legal character in an HTTP header.

So HTTP::Headers allows using a leading ":" for a header field name, which prevents this canonicalization.

e.g.
$h->header (':login_ldap' => 'some-value')

sets the "login_ldap" header.  Without the leading colon it would set the "Login-Ldap" header.

lwp-request has no way of specifying non-canonicalized headers.

Changing the regex in the split of the 'H' option fixes this, see attached patch
[non-canonicalized.txt](https://github.com/libwww-perl/libwww-perl/files/12737612/non-canonicalized.txt)

oalders:
- Stop testing Perls < 5.14 in CI
- Allow underscores not to be removed in lwp-request

Closes #442


karenetheridge:
is it possible to test this?

simbabque:
I've rebased this. I do agree tests would be nice, but that'll be a bigger job to write them. I've had a look and I think we don't have tests for any of the scripts, nor a dependency for running tests on them.

oalders:
I've cherry-picked this into `master`.

Issue #449, opened 2024-01-15
Issue with LWP::UserAgent, https IPv6 endpoint through proxy (using CONNECT): 400: Bad request


justnoxx:
Dear libwww-perl team!

Recently we faced an issue with reaching our backend by ipv6 address (https proto) through http proxy.

Our code looks like the following:
```perl
use strict;
use warnings;
use Data::Dumper;
use LWP::UserAgent;

my $ssl_tunnel3 = '[fd00:0:1:1f::4]:8443/commanderRequest';
my $proxy = 'http://localhost:6800';

my $ua = LWP::UserAgent->new();
$ua->proxy(['http', 'https'], $proxy);
my $req = HTTP::Request->new(
	'POST',
    "https://$ssl_tunnel2"
);


my $resp = $ua->request(
    $req
);

print Dumper $resp;
```

Returns 400: bad request.

After some investigation it turned out that `LWP::Protocol::http` in `request` function creates `$host:$port` pair using concatenation through `:`, they are:

https://github.com/libwww-perl/libwww-perl/blob/master/lib/LWP/Protocol/http.pm#L155
```perl
my $ssl_tunnel = $proxy && $url->scheme eq 'https'
	&& $url->host.":".$url->port;
```

https://github.com/libwww-perl/libwww-perl/blob/master/lib/LWP/Protocol/http.pm#L161
```perl
my $fullpath =
	$method eq 'CONNECT' ? $url->host . ":" . $url->port :
	$proxy && ! $ssl_tunnel ? $url->as_string :
	do {
	    my $path = $url->path_query;
	    $path = "/$path" if $path !~m{^/};
	    $path
	};
```

In this code the `$url` variable is an object of `URI::*` class. As per documentation of URI (https://metacpan.org/pod/URI#$uri-%3Ehost(-$new_host-) ) the following statement is present:
> For IPv6 addresses the brackets around the raw address is removed in the return value from $uri->host. When setting the host attribute to an IPv6 address you can use a raw address or one enclosed in brackets. The address needs to be enclosed in brackets if you want to pass in a new port value as well.

So when our ipv6 address, the `[fd00:0:1:1f::4]:8443` comes into one of previous code it is being parsed into `URI` object and then concatenated back to `fd00:0:1:1f::4:8443`. Which is incorrect, so we are getting error at `LWP::Protocol::http`, L201: https://github.com/libwww-perl/libwww-perl/blob/master/lib/LWP/Protocol/http.pm#L201

[...]

PR #450, opened 2024-01-22
Making it possible to use IPv6 in https call through https proxy environment (in case of using CONNECT method to create a tunnel)

justnoxx:
More details are at https://github.com/libwww-perl/libwww-perl/issues/449


haarg (approved):

oalders:
Thanks @justnoxx and @haarg!

PR #452, opened 2024-01-24
Update lwp-request to suport PATCH HTTP method

javierpuche:
Added PATCH as a supported method like POST and PUT
Deleted reference to CHECKIN method in help as it is not implemented


oalders (dismissed):
🥳

oalders (approved):
Thanks, @javierpuche! I will release this shortly.

javierpuche:
Thanks to you.

 When I or someone finds some spare time, maybe we can review if adding
the rest of not yet implemented methods from the official HTTP methods
list (https://www.w3.org/Protocols/HTTP/Methods.html) are as easy to
add as this one.

Regards.

oalders:
That page is a blast from the past! I think we could work from this document: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

PR #455, opened 2024-01-24
Simplify code slightly for Perl v5.8+

JRaspass:
Since LWP 6.00 in 2011 we require Perl v5.8.8 or better which makes these checks redundant.

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -775,7 +775,7 @@ sub parse_head {
                require HTML::HeadParser;
                $parser = HTML::HeadParser->new;
                $parser->xml_mode(1) if $response->content_is_xhtml;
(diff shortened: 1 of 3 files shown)


oalders (approved):
Thanks, @JRaspass!

lib/LWP/UserAgent.pm
```diff
@@ -775,7 +775,7 @@ sub parse_head {
                require HTML::HeadParser;
                $parser = HTML::HeadParser->new;
                $parser->xml_mode(1) if $response->content_is_xhtml;
-               $parser->utf8_mode(1) if $] >= 5.008 && $HTML::Parser::VERSION >= 3.40;
+               $parser->utf8_mode(1) if $HTML::Parser::VERSION >= 3.40;
```
oalders, line 778: Looks like 3.40 was released in 2004. 😅

PR #463, opened 2024-08-13
Fix proxy credentials with special characters

### LINKED ISSUE #456: Proxy basic auth failed if username starts with special symbol
Hi there,

I have faced with issue while tried configure connection through proxy server with basic authorization and my user contain special character as first symbol.

Code for reproducing:
```
#!/usr/bin/perl
use strict;
use warnings;
use v5.10;
use LWP::UserAgent;
use URL::Encode qw(url_encode);

my $proxy_user = url_encode('@dm!n_user');
my $proxy_pass = url_encode('$t0ng_p@$$w0rd!');
my $proxy_host = '172.19.18.200:8888';

my $ua = LWP::UserAgent->new();
$ua->proxy(["http", "http://$proxy_user:$proxy_pass\@$proxy_host"]);
my $result = $ua->get('http://ifconfig.me/ip');

say $result->status_line;
say $result->content;
``` 

Which cause next error:
```
Bad http proxy specification 'http://%40dm%21n_user:%24t0ng_p%40%24%24w0rd%21@172.19.18.200:8888' at test.pl line 13.
```

One more thing, I'm using url_encode method to escape special characters with their ASCII values, and when I specify proxy as array it doesn't decode back this symbol, but, when I specify as a hash, like next:
`$ua->proxy(http => "connect://$proxy_user:$proxy_pass\@$proxy_host");`
Symbols were decoded back which cause next behaviour:
```
500 Bad http proxy specification 'http://@dm!n_user:$t0ng_p@$$w0rd!@172.19.18.200:8888/'
Bad http proxy specification 'http://@dm!n_user:$t0ng_p@$$w0rd!@172.19.18.200:8888/' at /usr/share/perl5/LWP/Protocol/connect/Socket/Base.pm line 30.
```

As solution, I just allowed % symbol in proxy sub and defined proxy as array:
```
--- /usr/share/perl5/LWP/UserAgent.pm_orig      2024-02-05 11:02:39.204250417 -0500
+++ /usr/share/perl5/LWP/UserAgent.pm   2024-02-05 11:03:51.735193934 -0500
@@ -1079,7 +1079,7 @@
         my $url = shift;
         if (defined($url) && length($url)) {
             Carp::croak("Proxy must be specified as absolute URI; '$url' is not") unless $url =~ /^$URI::scheme_re:/;
-            Carp::croak("Bad http proxy specification '$url'") if $url =~ /^https?:/ && $url !~ m,^https?://[\w[],;
+            Carp::croak("Bad http proxy specification '$url'") if $url =~ /^https?:/ && $url !~ m,^https?://[\w[%],;
         }
         $self->{proxy}{$key} = $url;
         $self->set_my_handler("request_preprepare", \&_need_proxy)
```

However, I believe special charsets shouldn't be decoded back when we pass proxy details as hash and if username contain special character as first symbol it should be allowed.

kastakhov:
Fixes #456

```
me@rogflowx13:~/libwww-perl$ prove -l t/**/
t/base/default_content_type.t .. ok     
t/base/protocols.t ............. ok   
t/base/proxy.t ................. ok     
t/base/proxy_request.t ......... ok   
t/base/simple.t ................ ok   
t/base/ua.t .................... ok    
t/base/ua_handlers.t ........... ok   
t/leak/no_leak.t ............... skipped: Need Test::LeakTrace
t/local/autoload-get.t ......... ok   
t/local/autoload.t ............. ok   
t/local/cookie_jar.t ........... ok   
t/local/download_to_fh.t ....... ok   
t/local/get.t .................. ok   
t/local/http.t ................. ok       
t/local/httpsub.t .............. ok   
t/local/protosub.t ............. ok   
t/robot/ua-get.t ............... ok     
t/robot/ua.t ................... ok     
All tests successful.
Files=18, Tests=314, 26 wallclock secs ( 0.09 usr  0.01 sys +  1.67 cusr  0.39 csys =  2.16 CPU)
Result: PASS

me@rogflowx13:~/libwww-perl$ prove -lv t/base/proxy.t 
t/base/proxy.t .. 
1..20
ok 1 - proxy: with env: ABSURDLY_NAMED_PROXY: no errors
ok 2 - proxy: with env: MY_PROXY: no errors
ok 3 - HTTP_PROXY ignored in CGI environment
ok 4 - substitute CGI_HTTP_PROXY used in CGI environment
ok 5 - expected warning on multiple definitions
ok 6 - No warnings if multiple definitions for 'http_proxy' exist, but with the same value
ok 7 - no_proxy from environment
not ok 8 - No warnings for unrelated environment variables # TODO Test case for GH #372
#   Failed (TODO) test 'No warnings for unrelated environment variables'
#   at t/base/proxy.t line 83.
#     Structures begin differing at:
#          $got->[0] = 'Environment contains multiple differing definitions for 'foo'.
#     Using value from 'FOO' (BAR) and ignoring 'foo' (bar) at /home/me/libwww-perl/lib/LWP/UserAgent.pm line 1186.
#     '
#     $expected->[0] = Does not exist
ok 9 - http_proxy from environment
ok 10 - proxy: user with semicolon: got exception
ok 11 - proxy: password with semicolon: got exception
ok 12 - proxy: no host with env_proxy: got exception
ok 13 - http_proxy from method
ok 14 - proxy: user with semicolon: got exception
ok 15 - proxy: password with semicolon: got exception
ok 16 - proxy: no host: got exception
ok 17 - http_proxy from method
ok 18 - proxy: user with semicolon: got exception
ok 19 - proxy: password with semicolon: got exception
ok 20 - proxy: no host: got exception
ok
All tests successful.
Files=1, Tests=20,  0 wallclock secs ( 0.02 usr  0.00 sys +  0.07 cusr  0.02 csys =  0.11 CPU)
Result: PASS
```

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -1140,7 +1140,21 @@ sub proxy {
         my $url = shift;
         if (defined($url) && length($url)) {
             Carp::croak("Proxy must be specified as absolute URI; '$url' is not") unless $url =~ /^$URI::scheme_re:/;
-            Carp::croak("Bad http proxy specification '$url'") if $url =~ /^https?:/ && $url !~ m,^https?://[\w[],;
+            if ( $url =~ m!^(https?://)(.*)! ) {
+                my $scheme = $1;
+                my $path = $2;
+                if ($path =~ /@/) {
+                    my ($credentials, $host) = $path =~ /^(.+)@(.*)$/;
+                    Carp::croak("Bad http proxy specification with '$url'") unless $host;
+                    my ($user, $pass) = $credentials =~ /^(.*):(.*)$/;
+                    Carp::croak("Neither user nor password can contain ':' symbol") if $user =~ /:/;
+                    $user =~ s/([^\w])/sprintf("%%%02X", ord($1))/ge;
+                    $pass =~ s/([^\w])/sprintf("%%%02X", ord($1))/ge;
+                    $url = "${scheme}${user}:$pass\@$host";
+                } elsif ( $path !~ m,[\w[], ) {
+                    Carp::croak("Bad http proxy specification with '$url'");
+                }
+            }
         }
         $self->{proxy}{$key} = $url;
         $self->set_my_handler("request_preprepare", \&_need_proxy)
@@ -1165,7 +1179,7 @@ sub env_proxy {
             next if $k =~ /^HTTP_/;
             $k = "HTTP_PROXY" if $k eq "CGI_HTTP_PROXY";
         }
-	$k = lc($k);
+        $k = lc($k);
         if (my $from_key= $seen{$k}) {
             warn "Environment contains multiple differing definitions for '$k'.\n".
                  "Using value from '$from_key' ($ENV{$from_key}) and ignoring '$real_key' ($v)"
@@ -1175,18 +1189,17 @@ sub env_proxy {
             $seen{$k}= $real_key;
         }
 
-	next unless $k =~ /^(.*)_proxy$/;
-	$k = $1;
-	if ($k eq 'no') {
-	    $self->no_proxy(split(/\s*,\s*/, $v));
-	}
-	else {
+        next unless $k =~ /^(.*)_proxy$/;
+        $k = $1;
+        if ($k eq 'no') {
+            $self->no_proxy(split(/\s*,\s*/, $v));
+        } else {
             # Ignore random _proxy variables, allow only valid schemes
             next unless $k =~ /^$URI::scheme_re\z/;
             # Ignore xxx_proxy variables if xxx isn't a supported protocol
             next unless LWP::Protocol::implementor($k);
-	    $self->proxy($k, Encode::decode(locale => $v));
-	}
+            $self->proxy($k, Encode::decode(locale => $v));
+        }
     }
 }
 
@@ -1721,6 +1734,10 @@ Calling C<no_proxy> without any domains clears the list of domains.
         [ 'http', 'https' ] => 'http://http.example.com:8001/',
     ]);
 
+    # Proxy with authentication:
+    $ua->proxy('http', 'http://username:password@proxy.example.com/');
+    $ua->proxy('http' => 'http://username:password@proxy.example.com/');
+
 Set/retrieve proxy URL for a scheme.
 
 The first form specifies that the URL is to be used as a proxy for
@@ -1732,6 +1749,8 @@ proxy URL for a single access scheme.
 
 The third form demonstrates setting multiple proxies at once. This is also
 the only form accepted by the constructor.
+
+Note: if proxy credentials contain special characters, they will be encoded.
 
 =head1 HANDLERS
 
--- a/README.md
+++ b/README.md
@@ -493,6 +493,10 @@ Calling `no_proxy` without any domains clears the list of domains.
         [ 'http', 'https' ] => 'http://http.example.com:8001/',
     ]);
 
+    # Proxy with authentication:
+    $ua->proxy('http', 'http://username:password@proxy.example.com/');
+    $ua->proxy('http' => 'http://username:password@proxy.example.com/');
+
 Set/retrieve proxy URL for a scheme.
 
 The first form specifies that the URL is to be used as a proxy for
@@ -504,6 +508,8 @@ proxy URL for a single access scheme.
 
 The third form demonstrates setting multiple proxies at once. This is also
 the only form accepted by the constructor.
+
+Note: if proxy credentials contain special characters, they will be encoded.
 
 # HANDLERS
 
--- a/t/base/proxy.t
+++ b/t/base/proxy.t
@@ -4,7 +4,7 @@ use Test::More;
 use Test::Fatal qw( exception );
 
 use LWP::UserAgent ();
-plan tests => 8;
+plan tests => 20;
 
 # in case already defined in user's environment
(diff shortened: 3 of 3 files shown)


kastakhov:
Hi @oalders, actually it should be already good for review.
I had realized that forgot to cover one case and fixed it with my last change.
Also, added some additional tests just in case.

lib/LWP/UserAgent.pm
```diff
@@ -1140,7 +1140,24 @@ sub proxy {
...
+            my $host_re = '^[\w\[\]:.-]+/?$';
+            if ( $url =~ m!^(https?://)(.*)! ) {
+                my $scheme = $1;
+                my $path = $2;
+                if ($path =~ /@/) {
+                    my ($credentials, $host) = $path =~ /^(.+)@(.*)$/;
+                    Carp::croak("Bad http proxy specification with '$url'") unless $host && $host =~ m/$host_re/;
+                    my ($user, $pass) = $credentials =~ /^(.*):(.*)$/;
+                    Carp::croak("Neither user nor password can contain ':' symbol") if $user =~ /:/;
+                    $user =~ s/([^\w])/sprintf("%%%0x", ord($1))/ge;
```
genio, line 1154: Should we not be using [URI](https://metacpan.org/pod/URI) directly? If we don't want to cleanse that URI with URI directly, then maybe [URI::Escape::uri_escape_utf8()](https://metacpan.org/pod/URI::Escape#uri_escape_utf8(-$string-)) instead of redoing it ourselves here?

Just attempt to make a URI object instance, then grab the `->as_string` of it to have a properly dealt with URI string.

Apologies if I'm missing something, but it seems like we're adding yet another place for URI encoding rather than using the tools we already have a dependency on. If the URI module needs updating after testing this way, we should likely fix that instead, right?

kastakhov: Yeah, actually you're right, we can use URI / URI::Escape modules here.
Just one moment, should I load them globally or require inside proxy subroutine?
I see that in lib/LWP/Protocol/http.pm the URI::Escape module is loaded only when needed, I guess there should be the same behavior, am I right?

kastakhov:
Hmm, seems this code doesn't handle special characters in username/password. 
If credentials contain '@' symbol, they will be parsed incorrectly...
https://github.com/libwww-perl/libwww-perl/blob/b8d96f01ea318365693a43ded28109a79ed941bd/lib/LWP/Protocol/http.pm#L92
And as @oalders said correctly, credentials should not contain colon symbol, but there also no verification for this.

Will try to implement some common solution and use it in both places.

As for now, I'm converting this review to a draft.

lib/LWP/Protocol/http.pm
```diff
@@ -89,7 +89,7 @@ sub _fixup_header
 
     # Extract 'Host' header
     my $hhost = $url->authority;
-    if ($hhost =~ s/^([^\@]*)\@//) {  # get rid of potential "user:pass@"
+    if ($hhost =~ s/^(.*)@//) {  # get rid of potential "user:pass@"
```
oalders, line 92: This looks to me like the equivalent of:

Taking `$url->authority()` and then removing `$url->userinfo()`, but I'm not sure how that differs from just using `$url->host()`. That logic obviously predates this change. 🤔 

I tried to emit some debugging via the test suite, but it appears that there is no test coverage of when this condition is actually true.

kastakhov: Yep, you might be right (and I'm 99% sure you're).
I might work on it as well, to simplify the logic of this, but this is pretty old code without UT coverage I don't want to broke something here. 
And to be honest, I'm not sure how the UT should look like to cover this case...
Anyway, at least I'm 100% sure that now it'll catch the credentials correctly, when they contain "@" symbol.

kastakhov: For some reason request failed at this line if authorization contain "#" symbol:
https://github.com/libwww-perl/libwww-perl/blob/b8d96f01ea318365693a43ded28109a79ed941bd/lib/LWP/Protocol/http.pm#L214

URI lib cannot correctly parse next uri. For instance, next code:
```perl
use strict;
use warnings;
use v5.10;

use URI;

my $username = 'Str0ngUs3rW!th@#!$';
my $password = 'Str0ngP@$$w0rdW!th@#!$';
my $host = 'localhost';
my $port = '8080';

my $uri = URI->new("http://${username}:${password}\@${host}:${port}/path/to/file");
say $uri->scheme;
say $uri->host;
say $uri->port;
say $uri->path;
say $uri->authority;
say $uri->userinfo;
```
Return the next output:
```shell
$ perl test.pl
http

80

Str0ngUs3rW!th@
Str0ngUs3rW!th
```

I'd like to say it's a little broken...
Note: If I use uri_escape from URI::Escape before pass it into URI, everything worked as expected.
```shell
$ perl test.pl 
http
localhost
8080
/path/to/file
Str0ngUs3rW%21th%40%23%21%24:Str0ngP%40%24%24w0rdW%21th%40%23%21%24@localhost:8080
Str0ngUs3rW%21th%40%23%21%24:Str0ngP%40%24%24w0rdW%21th%40%23%21%24
```

I'll open a new issue in URI module....

oalders:
@kastakhov could you rebase this PR? I've added some codecov config which should (hopefully) add a coverage report to this PR.

lib/LWP/Protocol/http.pm
oalders: Nice find!

kastakhov:
Currently proxy.t will fail until https://github.com/libwww-perl/URI/issues/143 will be fixed...

Issue #309, opened 2019-03-14
Mention how to disable LWP::UserAgent timeout (via setting to 0 perhaps)


jidanni:
On LWP::UserAgent

Mention if setting this to 0 will
eliminate the timeout. Or mention how to.
```
   timeout
           my $secs = $ua->timeout;
           $ua->timeout( $secs );
```

oalders:
```perl
#!/usr/bin/env perl

use strict;
use warnings;
use feature qw( say );

use LWP::UserAgent;
my $ua = LWP::UserAgent->new;

say 'default timeout is: ' . $ua->timeout;

for my $timeout ( 1, 0, undef ) {
    my $ua  = LWP::UserAgent->new( timeout => $timeout );
    my $res = $ua->get('http://example.com');
    say 'timeout: ' . $ua->timeout;
    say $res->code;
    say $res->content if $res->code != 200;
}
```

```
default timeout is: 180
timeout: 1
200
timeout: 0
500
Can't connect to example.com:80 (Operation timed out)

Operation timed out at /Users/olaf/.plenv/versions/5.26.1/lib/perl5/site_perl/5.26.1/LWP/Protocol/http.pm line 50.

timeout: 180
200
```

jidanni:
I see.
Actually I don't see.
Anyway, maybe say on the man page / perldoc:

> To use a long timeout, try timeout=>9999999;

PR #466, opened 2024-11-27
Document that disabling the UA timeout is not supported

scop:
Closes https://github.com/libwww-perl/libwww-perl/issues/309


oalders (approved):

oalders:
Thanks, @scop!

PR #469, opened 2025-02-12
don't warn on proxy as unrecognized option for LWP::UserAgent->new

hillam:
Upon upgrading from perl 5.36.1 to perl 5.40.1, this error started showing up:
```
Unrecognized LWP::UserAgent options: proxy at /path/to/my/site_perl/5.40.1/LWP/UserAgent.pm line 102.
```

Sure enough, the `new` method seems to very clearly expect for `%cnf` to be empty before ever checking for `$cnf{proxy}`. I'm confused as to how nobody seems to have been hitting this previously.


oalders (approved):
Thanks, @hillam!

hillam:
@oalders do you have an idea of when I can expect to see this published to metacpan?

PR #476, opened 2025-06-27
Remove example references to malicious domain

dooleydevin:
[Some security vendors are considering `proxy[.]my[.]place` a malicious domain](https://www.virustotal.com/gui/domain/proxy.my.place/detection), causing `UserAgent.pm` to get flagged as potential malware by their scans. Updating the example documentation in `UserAgent.pm` and the `README` that includes this domain to avoid any confusion.


oalders (approved):
Thanks, @dooleydevin!

PR #478, opened 2025-07-10
[ci] Use perl-versions to build version matrix

happy-barney:
This way you don't need to update matrices every time new Perl version is released (like v5.42 is currently)

### DIFF
--- a/.github/workflows/dzil-build-and-test.yml
+++ b/.github/workflows/dzil-build-and-test.yml
@@ -70,7 +70,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        perl-versions: ${{ fromJson (needs.linux-perl-versions.outputs.perl-versions) }}
+        perl-version: ${{ fromJson (needs.linux-perl-versions.outputs.perl-versions) }}
     name: perl ${{ matrix.perl-version }}
     steps:
       - uses: actions/download-artifact@v4


oalders:
The build is broken because of an error in the workflow file.

oalders:
```
Error when evaluating 'strategy' for job 'linux-test-job'. .github/workflows/dzil-build-and-test.yml (Line: 73, Col: 24): Error from function 'fromJson': empty input, .github/workflows/dzil-build-and-test.yml (Line: 73, Col: 24): Unexpected value ''
```

@happy-barney did I clobber some of your changes?

happy-barney:
> ```
> Error when evaluating 'strategy' for job 'linux-test-job'. .github/workflows/dzil-build-and-test.yml (Line: 73, Col: 24): Error from function 'fromJson': empty input, .github/workflows/dzil-build-and-test.yml (Line: 73, Col: 24): Unexpected value ''
> ```
> 
> @happy-barney did I clobber some of your changes?

maybe I did, I did this error before, should be `fromJSON`, pushing change

oalders:
Looks like the Linux builds are missing the version name in the summary.

happy-barney:
> Looks like the Linux builds are missing the version name in the summary.

that's caused by mentioned missing `outputs.` ... instead of using output of job it got empty string

happy-barney:
I have to stop using `perl-version` and `perl-versions` ...

PR #501, opened 2026-03-24
Fix env_proxy() warning for unrelated environment variables

### LINKED ISSUE #372: environment case-insensitive warning painful

oalders:
Only warn about case-insensitive duplicate environment variables when
they are proxy-related (*_proxy). Previously, env_proxy() warned about
any duplicate env vars (e.g. FOO/foo), even on case-sensitive systems
and for variables LWP never reads.

The test coverage for this was added in GH #383 (v6.56) as a TODO test,
but the actual fix was never implemented. Remove the TODO marker now
that the test passes.

Based on a patch by @kberry.

Closes #372

### DIFF
--- a/lib/LWP/UserAgent.pm
+++ b/lib/LWP/UserAgent.pm
@@ -1169,9 +1169,12 @@ sub env_proxy {
         }
 	$k = lc($k);
         if (my $from_key= $seen{$k}) {
-            warn "Environment contains multiple differing definitions for '$k'.\n".
-                 "Using value from '$from_key' ($ENV{$from_key}) and ignoring '$real_key' ($v)"
-                if $v ne $ENV{$from_key};
+            # Only warn about proxy-related env vars, not unrelated ones (GH #372)
+            if ($k =~ /^(.*)_proxy$/) {
+                warn "Environment contains multiple differing definitions for '$k'.\n".
+                     "Using value from '$from_key' ($ENV{$from_key}) and ignoring '$real_key' ($v)"
+                    if $v ne $ENV{$from_key};
+            }
             next;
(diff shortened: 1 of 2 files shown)


oalders:
You're very welcome @kberry, but I only had 5 years to apply your patch. Imagine what I could have done had I been given more time!

thoke (approved):

oalders:
Thanks, everyone! libwww-perl-6.82.tar.gz has been uploaded to CPAN.

PR #505, opened 2026-05-08
Allow LWP::UserAgent to send headers on CONNECT requests

oalders:
Supersedes #92 (whose source fork was deleted, blocking further updates).

Rebased Mathias Fischer's original commits onto current `master` and resolved POD/test conflicts caused by the upstream switch from `=item` lists to `=head2` sections and from `plan tests => N` to `done_testing()`.

## Summary

- Adds `proxy_header($field [, $value])` to get/set individual headers for proxy CONNECT requests
- Adds `proxy_headers([$headers_obj])` to get/set the full `HTTP::Headers` object
- `LWP::Protocol::http` passes those headers on the `CONNECT` request when tunneling

This corresponds to curl's `--proxy-header` option.

Usage:

```perl
require LWP::UserAgent;

my $ua = LWP::UserAgent->new;
$ua->proxy(['http', 'https'], 'http://proxy.example.com');
$ua->proxy_header('User-Agent' => 'Special Agent');
```

## Test plan

- [x] `prove -lv t/base/ua.t` — all 53 tests pass locally
- [ ] CI green

### DIFF
--- a/Changes
+++ b/Changes
@@ -3,7 +3,8 @@ Change history for libwww-perl
 {{$NEXT}}
     - Add proxy_header() and proxy_headers() to LWP::UserAgent so that
       headers can be sent on the CONNECT request used to tunnel HTTPS
-      through an HTTP proxy (GH#505) (Mathias Fischer, Olaf Alders).
+      through an HTTP proxy. proxy_headers may also be passed to the
+      constructor (GH#505) (Mathias Fischer, Olaf Alders).
 
 6.82      2026-03-29 17:02:10Z
     - Fix env_proxy() warning for unrelated environment variables (GH#501)
--- a/Changes
+++ b/Changes
@@ -4,7 +4,7 @@ Change history for libwww-perl
     - Add proxy_header() and proxy_headers() to LWP::UserAgent so that
       headers can be sent on the CONNECT request used to tunnel HTTPS
       through an HTTP proxy. proxy_headers may also be passed to the
-      constructor (GH#505) (Mathias Fischer, Olaf Alders).
+      constructor (GH#505) (Mathias Fischer, Olaf Alders, Claude).
 
 6.82      2026-03-29 17:02:10Z
     - Fix env_proxy() warning for unrelated environment variables (GH#501)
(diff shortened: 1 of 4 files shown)


oalders:
## Code review — pass 3 (constructor support, `551945f5`)

Reviewing the new commit `551945f5` that adds `proxy_headers` as a constructor option.

### Verdict: Go

> "Verdict: GO. Ship it. … The asymmetry [vs `default_headers` wired directly into the bless] is fine and arguably better. Routing through the setter at construction time means a typo'd `proxy_headers => \$maybe_headers` fails immediately with a clear error rather than at first proxied request."

### Verified

- **Wiring**: setter has no dependencies on uninitialized state at the call site (between `from()` and `cookie_jar()`).
- **Validation**: `proxy_headers => 'garbage'` croaks at construction (better UX than `default_headers`, which silently stashes garbage and surfaces later).
- **Edge cases**: `undef`/`0`/`""` are all silent no-ops, consistent with the surrounding `if \$cookie_jar` / `if \$from` style.
- **POD example**: `HTTP::Headers->new('Proxy-Authorization' => ...)` is correct and idiomatic — matches the realistic CONNECT-time auth use case.
- **Ecosystem**: `LWP::RobotUA` and CPAN subclasses that `SUPER::new(@_)` get the feature transitively. Previously, `proxy_headers => ...` produced an "Unrecognized option" warning under `-w`; that warning now goes away. No breakage.
- **Reversing the 2017 "leave it out" decision** in the same release that introduces `proxy_header`/`proxy_headers` is cleaner than two separate change cycles.

### One non-blocking suggestion (addressed in `14b83e26`)

The reviewer suggested pinning the documented "stores the same object, not a copy" contract with an identity assertion. Added as commit `14b83e26`:

```perl
is($ua_ctor->proxy_headers, $ph,
    'constructor stores the same HTTP::Headers object (not a copy)');
```

### Status

- 64/64 tests pass in `t/base/ua.t`
- POD clean
- No regressions

Ready for the 6.83 release.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #531, opened 2026-08-13
Enforce max_size on reads that decode to no output

oalders:
## Summary

`LWP::UserAgent->max_size` could be bypassed entirely by a server whose
response body decodes to nothing under a Transfer-Encoding.

`LWP::Protocol::http::request` reads the body in a `collect()` callback.
When a Transfer-Encoding transform consumes input but yields no decoded
output, `Net::HTTP::read_entity_body` returns `-1`; the callback replied
with `redo READ` and never returned to `collect()`, where `max_size` is
checked. A peer can hold that loop open indefinitely — e.g. a
`Transfer-Encoding: deflate` stream of zlib sync-flush markers, each
inflating to zero bytes — reading unbounded data regardless of the cap.

## Fix

Track undelivered bytes in the callback and charge them against
`max_size` there. Decoded output is never smaller than the compressed
input it came from, so a transfer whose consumed bytes already exceed the
cap cannot decode to anything under it — aborting with
`Client-Aborted: max_size` is safe. The counter resets on any read that
delivers bytes, so normal compressed responses see no change in behaviour.

## Tests

- New `/syncflush` route in `t/local/http.t` serves a deflate/chunked
  stream of sync-flush markers; asserts a small `max_size` now aborts
  with `Client-Aborted: max_size`.
- File switched from a fixed `plan tests` to `done_testing()`.

All 138 subtests in `t/local/http.t` pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,10 @@
 Change history for libwww-perl
 
 {{$NEXT}}
+    - LWP::Protocol::http now enforces max_size against reads whose
+      Transfer-Encoding decodes to no output. Previously such reads bypassed
+      the size limit entirely, letting a server stream unbounded data under
+      any max_size.
 
 6.83      2026-05-12 11:41:48Z
     - LWP::UserAgent now strips Authorization and Proxy-Authorization headers
--- a/lib/LWP/Protocol/http.pm
+++ b/lib/LWP/Protocol/http.pm
@@ -458,6 +458,8 @@ sub request
     $response->push_header('Client-Response-Num', scalar $socket->increment_response_count);
 
     my $complete;
+    my $max_size = $self->{max_size};
+    my $undelivered = 0;
     $response = $self->collect($arg, $response, sub {
 	my $buf = ""; #prevent use of uninitialized value in SSLeay.xs
 	my $n;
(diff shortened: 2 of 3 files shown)


oalders (commented):
## Code review — `max_size` vs. zero-output Transfer-Encoding

Multi-lens review (general + security + Perl). Net assessment: **this is a genuine security improvement** — it converts a previously *unbounded* read-loop hang into a bounded one, with a clean abort path (`return \ ''` terminates `collect()`, `Client-Aborted: max_size` pushed, `$complete` left false so the socket is dropped rather than reused), a static header string (no info leak), and a regression test that genuinely fails without the patch. No Critical issues; nothing here is a merge blocker.

Three **Important** observations are inline below. Two of them are coupled: the tempting fix for the counter reset is unsafe *while* the per-read charge stays coarse — see the notes. I've already applied local fixes for the comment-accuracy and test-strength items (inline notes say which); the counter-reset item is left as a design question for you.

**Minor (informational):** for a `Transfer-Encoding: gzip` response, the terminal chunk is decompressed in one shot inside `Net::HTTP`'s `read_entity_body` and returned as a single large positive `$n`, which bypasses the new `-1` guard entirely and is only size-checked by `collect()` *after* the allocation — so a bounded compressed buffer can still spike memory/CPU. That path is a pre-existing `Net::HTTP` limitation, outside this diff's control; the **deflate** path this PR targets is not affected (it inflates incrementally). Worth a separate upstream note, not a change here.

**Strengths:** boundary (`> $max_size`, strict) matches `collect()`'s existing convention; `max_size` undef/0 and the EINTR/EWOULDBLOCK redo all handled correctly; accounting direction is fail-closed; `SKIP` correctly gates on `Compress::Raw::Zlib`; `done_testing()` is the right call over a brittle fixed count.

---
🤖 Review by [Claude Code](https://claude.com/claude-code) · Opus 4.8

robrwo (approved):

